PR #28, opened 2016-02-11
DOC-233: Using other storage engines

azhebel:


source/perconaft.rst
```diff
@@ -43,41 +75,7 @@ By default, Percona Server for MongoDB runs with MMAPv1. To enable PerconaFT:
...
-     - NO
-     - YES
-     - YES
-
-.. rubric:: Footnotes
+Using Other Storage Engines
+===========================
 
-.. [#n-1] This feature is currently considered *Experimental*
+|Percona Server for MongoDB| includes several storage engines, which have a different set of features and advantages over one another. You can switch between storage engines using the ``--storageEngine`` option. Also specify the ``--dbpath`` option to a separate data directory for every storage engine you use.
```
hrvojem: DOC-233 bug refers to post-install steps, due to changes in package installation script.

With new packaging once you install packages, server will automatically start after installation with the MMAPv1 as the default storage engine and if you try to change the engine simply with --storage-engine option it would fail.

Note should be provided here and more details in the deb installation guide on how to change the default storage engine after packages have been installed. (stop service, rm -rf mongod-datadir, change the storage engine in the mongo.conf, start service)

Docs are missing information on ability to change the storage engine via mongod.conf configuration file.

PR #33, opened 2016-04-05
DOC-246: Documenting YAML config options for PerconaFT

azhebel:


azhebel:
This has the wrong base branch. Here is the correct PR for 3.0
https://github.com/percona/psmdb-docs/pull/35

PR #48, opened 2016-09-16
Release notes for 3.2.9-2.1

azhebel:


dbadapt (changes requested):
Looks good!    Only think I noticed is that a period is needed after 2016.

PR #52, opened 2016-12-06
Release notes for 3.2.11-3.1

azhebel:


tplavcic:
The release notes should have info that we are adding ubuntu yakkety packages. Also we didn't have binary tarballs for ubuntu xenial which were added also.
https://jira.percona.com/browse/BLD-556 (internal task no link needed)

This should be also added to release notes (HotBackup functionality was missing from ubuntu/debian packages): https://jira.percona.com/browse/PSMDB-92

source/release_notes/3.2.11-3.1.rst
```diff
@@ -0,0 +1,31 @@
...
+Percona Server for MongoDB requires no changes to MongoDB applications or code.
+
+.. note:: The PerconaFT storage engine has been deprecated
+   and will not be available in future releases.
+
+This release is based on `MongoDB 3.2.11
+<http://docs.mongodb.org/manual/release-notes/3.2/#nov-18-2016>`_
+and includes the following fix related to PerconaFT:
+
+* Fixed crash during shutdown of ``mongod``
```
denis-protivenskii: Fixed hang

tplavcic:
@azhebel I had a discussion with @denis-protyvenskyi and we concluded that every bugfix and new feature should have a public ticket and they should be referenced in the release notes.
So for this "Added :ref:`hot-backup` to Ubuntu/Debian packages" I would like us to have a link to https://jira.percona.com/browse/PSMDB-92.
The other one (hang issue) currently doesn't have a ticket so unless Denis now want to open one let's leave it as it is.

tplavcic:
Also I might add that Ubuntu Wily is removed from supported platforms as that distribution has reached end of life so you might want to put that somewhere.

hrvojem:
Add note that the default configuration file has been changed/updated to include some of the options.

PR #59, opened 2017-05-08
Release notes for 3.4.4-1.4

azhebel:


source/release_notes/3.4.4-1.4.rst
```diff
@@ -0,0 +1,27 @@
...
+:Installation: :ref:`install`
+
+This release is based on `MongoDB 3.4.4
+<https://docs.mongodb.com/manual/release-notes/3.4/#apr-21-2017>`_,
+and includes the following additional changes:
+
+* :jirabug:`PSMDB-122`: Added the ``percona-server-mongodb-enable-auth.sh``
+  script to :ref:`binary tarball <tarball>`.
+
+* :jirabug:`PSMDB-127`: Fixed cleanup of deleted documents and indexes
```
igorsol, line 18: I discussed PSMDB-127 release notes with Peter Schwaller few weeks ago. And he proposed to put something in the release notes about "deferred compactions may occur when starting up with the new version". That also means that in some cases db size can decrease significantly after deferred compactions finish.

igorsol (changes requested):

source/release_notes/3.4.4-1.4.rst
azhebel: added in separate commit

PR #62, opened 2017-06-15
DOC-327: Added considerations for PSMDB storage engine configuration …

azhebel:
…templates


tplavcic (changes requested):
default cache size for rocksdb has changed, see DOC-440

PR #71, opened 2017-07-26
Update .gitignore

Oneiroi:
Add .DS_Store to the gitignore file, avoiding commits such as 79ce0a379d516fabf8739cc69fe507bcf5928b6a where the file is removed instead these Metafile should be ignored and not added at all.


azhebel:
Thank you!

PR #82, opened 2018-03-28
Fix for DOC-727, documenting deprecation of MongoRocks

fiowro:


source/mongorocks.rst
```diff
@@ -8,6 +8,12 @@ MongoRocks
 based on the RocksDB_ key-value store optimized for fast storage.
 It is developed by Facebook and designed to handle write-intensive workloads.
 
+.. warning:: *MongoRocks* is deprecated in 3.6 and will be removed in the
+   next major version! Feature compatibility version for MongoRocks
+   is limited to 3.4 so 3.6 functionality cannot be used. Read/write
+   concern majority is not safe to use as there are a number of
+   failing tests for those.
```
RoelVdP: Please change to;
is limited to 3.4, so 3.6 functionality cannot be used. Read/write concern majority is not safe to use.

source/mongorocks.rst
```diff
@@ -21,10 +27,11 @@ Using MongoRocks
 ================
 
 As of version 3.2, |PSMDB| runs with `WiredTiger`_ by default.
-You can select a storage engine
-using the ``--storageEngine`` command-line option when you start ``mongod``.
-Alternatively, you can set the ``storage.engine`` variable
-in the configuration file (by default, :file:`/etc/mongod.conf`):
+You can select a storage engine using the ``--storageEngine``
```
RoelVdP: This needs a lead-in. For example "If you still would like to use the deprecated MongoRocks storage engine, please use the ``--storageEngine rocksdb`` command-line option, accompanied by ...."

source/mongorocks.rst
```diff
@@ -21,10 +27,11 @@ Using MongoRocks
...
+command-line option accompanied by ``--useDeprecatedMongoRocks`` when
+you start ``mongod``. Alternatively, you can set the 
+``storage.engine`` and ``useDeprecatedMongoRocks`` variables in the
+configuration file (by default, :file:`/etc/mongod.conf`).
```
RoelVdP: ... configuration file (by default, :file:`/etc/mongod.conf`) as shown below.

RoelVdP (changes requested):
@tplavcic would you please do an all-over tech review? Including the final page as-is. Thanks

tplavcic:
This line in "Using MongoRocks" is not working: `sed -i '/engine: \*rocksdb/s/#//g' /etc/mongod.conf` from what I see it should be `sed -i '/engine: *rocksdb/s/#//g' /etc/mongod.conf`.
And since useDeprecatedMongoRocks is needed now I think this line should be added after that one: `sed -i '/engine: *rocksdb/a\  useDeprecatedMongoRocks: true' /etc/mongod.conf`

By looking in github I don't see that the warning of deprecation is boxed and colored but I think it will be on our web page if that is true then it's ok.

Should we add `useDeprecatedMongoRocks` in the options list? Most probably yes, although it's a mandatory switch so I'm not 100% on this one.

No other issues I see.

fiowro:
And what do you think about removing these sed lines from the 3.6 doc? They are just to uncomment "engine: rocksdb" line (which is commented in the example mongod.conf). Adding “useDeprecatedMongoRocks” by sed would either be capable to produce duplicate lines in config or would involve some complicated and unreadable sed magic… 

Necessary modifications to config file are already described, and later there is an example, so removing sed commands would probably do nothing wrong...

tplavcic:
@fiowro Removing those sed lines from docs works for me.

tplavcic:
LGTM, @RoelVdP if no additional comments can you please approve PR so that it can be merged. Thanks!

tplavcic:
@fiowro found one small issue on upgrade page, the step:
```#. Add ``useDeprecatedMongoRocks: true`` line into the config file.```
should only apply if the user is using MongoRocks, so maybe something like:
```#. Add ``useDeprecatedMongoRocks: true`` line into the config file if using MongoRocks storage engine.```

**Review by tplavcic — CHANGES_REQUESTED** [review]
As already discussed one small change is still needed on the upgrade page and then it can be merged.
#. Add ``useDeprecatedMongoRocks: true`` line into the config file.
should only apply if the user is using MongoRocks, so maybe something like:
#. Add ``useDeprecatedMongoRocks: true`` line into the config file if using MongoRocks storage engine.

File: source/install/upgrade_from_34.rst [hunk 4]
```diff
@@ -41,6 +44,11 @@ Upgrading on Debian or Ubuntu
 
     sudo apt-get remove percona-server-mongodb-34*
 
+#. Optionally add ``useDeprecatedMongoRocks: true`` line into the config
+   file if using MongoRocks storage engine::
+
+    sed -i '/engine: *rocksdb/a\ useDeprecatedMongoRocks: true' /etc/mongod.conf
```
RoelVdP: `sed -i '/engine:[ \t]*rocksdb/a\ useDeprecatedMongoRocks: true' /etc/mongod.conf`

Is safer as it includes tabs besides spaces

source/install/upgrade_from_34.rst
```diff
@@ -62,6 +70,11 @@ Upgrading on RHEL and CentOS
 
     sudo yum remove Percona-Server-MongoDB-34*
 
+#. Optionally add ``useDeprecatedMongoRocks: true`` line into the config
+   file if using MongoRocks storage engine::
+
+    sed -i '/engine: *rocksdb/a\ useDeprecatedMongoRocks: true' /etc/mongod.conf
```
RoelVdP: As above

RoelVdP (changes requested):

source/install/upgrade_from_34.rst
RoelVdP: Good to merge otherwise

RoelVdP (approved):

tplavcic (approved):
LGTM

peterschwaller:
Dummy comment to trigger sync to Jira.  Please ignore.

PR #108, opened 2018-11-01
DOC-783: Update based on feedback

borys-belinsky:


tplavcic (changes requested):
The box around: 
"Important
This feature is considered BETA quality. Do not use the data encryption at rest in a production environment." 
is still missing.

PR #120, opened 2019-02-05
PSMDB-314: perconadecrypt - add rubric

borys-belinsky:


source/data_at_rest_encryption.rst
```diff
@@ -25,6 +25,27 @@ services.
    This feature is considered **BETA** quality. Do not use the |feature| in a
```
tplavcic: I don't think this is BETA. From what I know the whole encryption is becoming GA.

tplavcic (changes requested):

source/data_at_rest_encryption.rst
```diff
@@ -20,10 +20,26 @@ The |feature| in |PSMDB| is introduced in version 3.6 to be compatible with
...
 
-   This feature is considered **BETA** quality. Do not use the |feature| in a
-   production environment.
+Starting from version 3.6, |PSMDB| supports the encryption of rollback files. To
+inspect the contents of these files, use |perconadecrypt|. This is a
+tool that you run from the command line as follows:
+
+.. code-block:: bash
+
+   $ perconadecrypt --encryptionKeyFile ... --encryptionCipherMode ... --inputPath --outputPath
```
tplavcic: "--inputPath" and "--outputPath" also take parameters so consider adding "..." after those options also.

source/data_at_rest_encryption.rst
```diff
@@ -20,10 +20,26 @@ The |feature| in |PSMDB| is introduced in version 3.6 to be compatible with
...
+
+.. admonition:: Parameters of |perconadecrypt|
+
+   ========================  ===================================================================
+   Option                    Purpose
+   ========================  ===================================================================
+   --encryptionKeyFile       The path to the encryption file
```
tplavcic: Add key file to this "The path to the encryption file" so it becomes "The path to the encryption key file"

source/data_at_rest_encryption.rst
```diff
@@ -20,10 +20,26 @@ The |feature| in |PSMDB| is introduced in version 3.6 to be compatible with
 not include support for |abbr.kmip|, |vault| or |amazon-aws| key management
 services.
 
-.. important:: 
+.. rubric:: Encrypting Rollback Files
 
-   This feature is considered **BETA** quality. Do not use the |feature| in a
-   production environment.
+Starting from version 3.6, |PSMDB| supports the encryption of rollback files. To
```
tplavcic: I wouldn't say that PSMDB supports the encryption of rollback files, but instead that "rollback files are encrypted if data at rest encryption is enabled." because there's no separate option for encryption of those files, either everything is encrypted together with rollback files or nothing is encrypted - we just missed to encrypt rollback files in the beta version so we added this functionality in the release after.

source/data_at_rest_encryption.rst
tplavcic: Maybe best use the format similar to what the tool is printing in the help screen: `perconadecrypt --encryptionKeyFile <key path> --encryptionCipherMode <cypher mode> --inputPath <src> --outputPath <dest>`

source/data_at_rest_encryption.rst
```diff
@@ -20,10 +20,26 @@ The |feature| in |PSMDB| is introduced in version 3.6 to be compatible with
...
+
+   $ perconadecrypt --encryptionKeyFile ... --encryptionCipherMode ... --inputPath --outputPath
+
+.. admonition:: Parameters of |perconadecrypt|
+
+   ========================  ===================================================================
+   Option                    Purpose
+   ========================  ===================================================================
+   --encryptionKeyFile       The path to the encryption file
+   --encryptionCipherMode    The cipher mode. The supported values are AES256-CBC or AES256-GCM
```
tplavcic: "The cipher mode." -> "The cipher mode to use for decryption."

tplavcic (changes requested):

source/data_at_rest_encryption.rst
borys-belinsky: Ready for review

tplavcic (approved):
LGTM

PR #121, opened 2019-02-06
PSMDB-320: Create section for ngram fts

borys-belinsky:


source/ngram-full-text-search.rst
```diff
@@ -0,0 +1,37 @@
...
+The |ngram| full text search is useful for searching text for a
+specific string of characters in a field of a collection. To use |ngram|, create
+a text index on a collection setting the ``default_language``
+parameter to **ngram**:
+
+.. code-block:: guess
+
+   mongo > db.collection.createIndex({name:"text"}, {default_language: "ngram"})
+
+The |ngram| search engine uses the following characters that not used as parts
```
tplavcic: This is strange and not sure what it means. Probably better would be "The ngram search engine uses the following characters as tokenizer delimiters:"

tplavcic (changes requested):
Based on the information here: https://github.com/percona/percona-server-mongodb/pull/365 (and from what I've seen in testing) I would also add:
- that this feature is useful for Korean, Chinese, Japanese languages
- that this feature is slower than normal MongoDB full text search

source/ngram-full-text-search.rst
borys-belinsky: Thanks for noting this problem. 'tokenizer delimiter' (tokenizer usually represents a tool which extracts tokens) sounds suspicious and i didn't find it used in reliable contexts. Using 'token delimiter character'.

source/ngram-full-text-search.rst
```diff
@@ -0,0 +1,40 @@
...
+
+The |ngram| full text search is useful for searching text for a specific string
+of characters in a field of a collection. To use |ngram|, create a text index on
+a collection setting the ``default_language`` parameter to **ngram**:
+
+.. code-block:: guess
+
+   mongo > db.collection.createIndex({name:"text"}, {default_language: "ngram"})
+
+The |ngram| search engine uses the following token delimiter characters that not
```
tplavcic: "that not used" -> "that are not used"

tplavcic (approved):
Commented one small change, otherwise ok.

igorsol (approved):

PR #126, opened 2019-02-25
PSMDB-329: add relaxPermChecks option

borys-belinsky:


tplavcic:
This looks ok to me, except that this option is not only used for data at rest encryption key options but also for keys for internal auth between nodes in which case I'm not sure where to put this info. @igorsol can you please comment!

tplavcic:
What I wanted to say is that this option affects how the permissions of the file specified under `--keyFile` option are handled, not just the one specified with `--encryptionKeyFile`

source/data_at_rest_encryption.rst
```diff
@@ -83,12 +83,21 @@ key and save it to a file by using the |openssl| command:
...
+specified with the ``chmod`` command can either be **600** (only the owner may
+read and modify the file) or **400** (only the owner may read the file.)
 
 .. code-block:: bash
 
    $ chmod 600 mongodb-keyfile
 
+If ``mongod`` is started with the ``--relaxPermChecks`` option, the keyring file
+can be owned by ``root`` and ``mongod`` can read or change the file based on the
+group bit set accordingly. The effective keyring file permissions in this
```
igorsol: Please consider this variant:
```
If ``mongod`` is started with the ``--relaxPermChecks`` option and the keyring file
is owned by ``root`` then ``mongod`` can read the file based on the
group read bit set accordingly.
```

igorsol: This part: `mongod can read or change` should be `mongod can read` because:
1. mongod won't try to change key file
2. if group `write` bit is set mongod won't start

igorsol (changes requested):

source/data_at_rest_encryption.rst
borys-belinsky: Ready for review

tplavcic (approved):
LGTM

igorsol (approved):

PR #136, opened 2019-06-04
PSMDB-373-update-docs-hashicorp-vault-support updated text and variants

patrickbirch:
Updated text based on Igor's comments.


igorsol (approved):

patrickbirch:
Added a commit that clarifies the support for V2 API versioning must be enabled.

tplavcic (changes requested):
I don't see key rotation steps?

patrickbirch:
Added rotation steps.

tplavcic (approved):
LGTM

PR #143, opened 2019-07-23
PSMDB_406: vault encryption - add migration steps

borys-belinsky:


source/data_at_rest_encryption.rst
```diff
@@ -8,6 +8,10 @@ Data at rest encryption for the WiredTiger storage engine in |mongodb| was
 introduced in |mongodb-enterprise| version 3.2. to ensure that encrypted data
 files can be decrypted and read by parties with the decryption key.
 
+Note that data at rest encryption is enabled when starting |PSMDB| providing all
+essential encryption options. You cannot enable or disable encryption while the
+|PSMDB| server is already running.
```
tplavcic: Not only that, you cannot change the options even if you stop the server - eg. if it was started with some encryption options it should always be started the same, so even if you stop it you cannot switch from encrypted->non-encrypted or vice versa.

source/data_at_rest_encryption.rst
```diff
@@ -77,10 +54,10 @@ the encryption. By default, the |opt.encryption-cipher-mode| option uses the
    --outputPath              The path to save the decrypted rollback file
    ========================  ==================================================================================
 
-.. rubric:: Important Configuration Options
+.. rubric:: Important Configuration Options When Using Key File Encryption
```
tplavcic, line 61: `--enableEncryption` and `--encryptionCipherMode` are used for both keyfile and vault, so they are common options. Others depends where you store the key.

tplavcic (changes requested):

source/data_at_rest_encryption.rst
```diff
@@ -17,44 +21,17 @@ files can be decrypted and read by parties with the decryption key.
...
-
-On start, the server tries to read the master key from the Vault. If the
-configured secret does not exist, Vault responds with the HTTP 404 error. During
-the first run of |PSMDB|, the process generates a secure key and writes the key
-to the vault.
-
-.. rubric:: Encrypting Rollback Files
-
-Starting from version 3.6, |PSMDB| also encrypts rollback files when data at
+Starting from version 3.6, |PSMDB| enables encrypting rollback files when data at
```
igorsol: This was added in version 3.6.10-3.0 as a fix for PSMDB-273

source/data_at_rest_encryption.rst
```diff
@@ -17,44 +21,17 @@ files can be decrypted and read by parties with the decryption key.
 
 The |feature| in |PSMDB| is introduced in version 3.6 to be compatible with
```
igorsol: encryption was beta in 3.6.8-2.0 and GA in 3.6.10-3.0

igorsol (changes requested):

source/data_at_rest_encryption.rst
```diff
@@ -150,6 +132,54 @@ All these options can be specified in the configuration file:
...
+You can use either the |vault| integration feature or the key file
+encryption. If you have used the key file encryption and wish to migrate to  
+using |vault|, complete the following steps:
+
+1. Insert the key from keyfile into the |vault| server to the desired secret
+   path.
+#. Stop mongod.
+#. Add the startup options for |vault| integration for data at rest encryption.
+#. Remove the startup options for the key file encryption.
+#. Start the ``mongod`` service
```
igorsol, line 181: Well, this procedure should work but I think it would be great to provide some examples along the way. How to insert key into the vault? Which parameters should be added for vault and removed for keyfile?

igorsol (changes requested):

source/data_at_rest_encryption.rst
```diff
@@ -77,10 +58,10 @@ the encryption. By default, the |opt.encryption-cipher-mode| option uses the
...
-.. rubric:: Important Configuration Options
+.. rubric:: Important Configuration Options When Using Key File Encryption
 
-|PSMDB| supports the ``encryptionCipherMode`` option where you choose one of the
-following cipher modes:
+With encryption key files, |PSMDB| supports the ``encryptionCipherMode`` option
+where you choose one of the following cipher modes:
 
 - |mode.cbc|
 - |mode.gcm|
```
igorsol, line 67: `encryptionCypherMode` is supported regardless where the encryption key is stored

nastena1606:
Reworked in https://github.com/percona/psmdb-docs/pull/270

PR #185, opened 2020-04-06
PSMDB-386 Updated previous version variable, added new step to upgrad…

nastena1606:
…e checklists

### DIFF
--- a/source/.res/replace.txt
+++ b/source/.res/replace.txt
@@ -3,7 +3,7 @@
 .. |mongodb-ce| replace:: MongoDB 4.0 Community Edition
 .. |mongodb| replace:: MongoDB
 .. |percona| replace:: Percona
-.. |prev-version| replace:: 4.0
+.. |prev-version| replace:: 3.6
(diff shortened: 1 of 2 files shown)


tplavcic (approved):
LGTM

source/.res/replace.txt
```diff
@@ -3,7 +3,7 @@
 .. |mongodb-ce| replace:: MongoDB 4.0 Community Edition
 .. |mongodb| replace:: MongoDB
 .. |percona| replace:: Percona
-.. |prev-version| replace:: 4.0
+.. |prev-version| replace:: 3.6
```
tplavcic, line 6: I'm not 100% sure that I know what this does, but are you sure 4.0 should be replaced with 3.6 here - notice you are mergeing to 4.0 branch?

nastena1606: There was a bug with previous version variable (from 4.0 to 4.0). I fixed that.

tplavcic: Cool! Thanks!

PR #195, opened 2020-04-22
PSMDB-603 Fix update percona-release command for .rpm distros

nastena1606:


source/.res/text/important.percona-release.latest.yum.txt
```diff
@@ -0,0 +1,9 @@
+
+.. important::
+
+   If |percona-release| is already installed on the target system, make sure
+   you are using the latest version of |percona-release|:
+
+   .. code-block:: bash
+
+      $ sudo yum update percona-release
```
EvgeniyPatlan, line 9: this command will work only if original repo is enabled

nastena1606: So the command must be preceded with the command:
$ sudo percona-release enable original 
Correct?

I believe the same is valid for DEB-based systems

EvgeniyPatlan: @nastena1606 this is more complicated question. for example the might be package conflict while having orignal repo and some other repos.
another solution is to download latest percona-release package from web and install it.
for rpms it can be yum install <url-to_latest.rpm>
for debs - download and install

nastena1606: But this is the thing. The warning states that if the user has percona-release already installed, there's the command to check for the latest version. Unfortunately there's no command (documented at least) to check what percona-release is installed / enabled (https://www.percona.com/doc/percona-repo-config/percona-release.html#commands)

If the solution is to download the latest percona-release package and this resolves package conflicts and/or disables unrequired repos to ensure the correct one is used, then there's no need in this warning at all.

borys-belinsky (approved):

EvgeniyPatlan (commented):
lgtm :)

PR #214, opened 2020-05-21
PSMDB-629-Release notes for 3.6.18-6.0

nastena1606:


source/release_notes/3.6.18-6.0.rst
```diff
@@ -0,0 +1,32 @@
...
+================================================================================
+
+* :jirabug:`PSMDB-616`: Add "txn_checkpoint_prepare_time" and "txn_checkpoint_tree_helper_time" wiredTiger.transaction stats
+
+
+
+Improvements
+================================================================================
+
+* :jirabug:`PSMDB-587`: Add options to control WiredTiger data-handle hash size
```
igorsol: I think this should be under "new features" too.
Since we added several options to control sizes of several arrays (not only data-handle array) I think it would be better to write `Add options to control size of various wiredTiger's hash arrays`.
Let's wait for an opinion from @akira-kurogane

akira-kurogane: Hi Anastasia and Igor.

I would describe this patch as adding options to control the size of WiredTiger's datahandle hash arrays.

So: option, size of hash arrays, datahandles are the things being hashed.

What effect this option has: This is a tuning option that may remediate checkpoint latency caused by WiredTiger datahandle sweeps. Note: This tuning option is only intended for mongod nodes with greater than 100,000 WiredTiger data and index files.

akira-kurogane: And yes, and although this is an improvement, it's even more suitable i.m.o. to categorize it as a new feature.

akira-kurogane (changes requested):

source/release_notes/3.6.18-6.0.rst
igorsol: @akira-kurogane I'm not sure only datahandles are hashed in those arrays. We have three new parameters:
1. session_dhhash_size - this one is for datahandles
2. session_cursor_cache_size - size of hash array that holds cached session cursors
3. big_hash_array_size - this one is used to specify size of three connection level arrays:
3.1. dhhash (data handle hash array)
3.2. fhhash (commented as 'file list' seems to be acronym for 'file handle hash array')
3.3. blockhash (commented as 'block manager list')

akira-kurogane: Okay, fair criticism.

In my defense I became accustomed to calling this the 'datahandles' patch for a few reasons - Netskope's contributors do; the metrics I watch in FTDC / serverStatus are about the datahandle sweeps; and as I recall datahandles are one-to-one with filehandles, at least the way MongoDB uses WiredTiger library.

There would be other hash arrays I would bet, so I don't like to say just 'various wiredTiger's hash arrays'.

How about we say 'Adding options to control the size of several WiredTiger hash arrays concerned with datahandles and one for session cursor cache'. (I think the blockhash is not important to differentiate; it moves with the other two in #3, but doesn't seem to be a performance influencer.)

igorsol: works for me

nastena1606: Akira and Igor, thanks for the review. For the issue to appear in "New features" section after release-notes auto-generation, I must change its type to New Feature in Jira.

akira-kurogane (approved):
Thank you Anastasia!

igorsol (approved):
LGTM

PR #220, opened 2020-06-09
PSMDB-546 Added restore mongod permissions to custom paths step

nastena1606:


sandraromanchenko (changes requested):
1. No need to check mongod.conf for Debian/Ubuntu, by default, file isn't replaced by newer version from package during upgrade
2. No section with link to https://docs.mongodb.com/ upgrade manual
3. Maybe it's worth replacing command `service mongod start/stop` with `systemctl start/stop mongod` as command is quite outdated

sandraromanchenko (changes requested):
Command systemctl should look like 'systemctl stop mongod' and it should be the same for Debian/Ubuntu and RHEL/Centos, the remaining LTGM.

sandraromanchenko (approved):

PR #237, opened 2020-07-13
PSMDB-228 Add upgrade replica set steps

nastena1606:


source/install/upgrade_from_mongodb.rst
```diff
@@ -17,38 +18,46 @@ An in-place upgrade is done with existing data in the server.  Generally
 speaking, this is stopping the server, removing the old packages, installing the
 new server and starting it with the same data files.  While an in-place upgrade
 may not be suitable for high-complexity environments, it should work in most
-cases.
+cases.  
 
 .. warning::
 
-   Before starting the upgrade process it is recommended to perform a full
+   Before starting the upgrade service, we recommend to perform a full
```
igorsol: I guess the 'upgrade process' was used with meaning of 'upgrade procedure' here. So it seems 'service' is not applicable here

source/install/upgrade_from_mongodb.rst
```diff
@@ -17,38 +18,46 @@ An in-place upgrade is done with existing data in the server.  Generally
...
 .. contents::
    :local:
 
-Upgrading on Debian or Ubuntu
+.. _upgrade_standalone:   
+
+Upgrading a standalone instance or a single-node replica set 
 ================================================================================
 
+The upgrade service depends on the distribution you are using:
```
igorsol: upgrade procedure?

source/install/upgrade_from_mongodb.rst
```diff
@@ -100,5 +117,35 @@ Upgrading on Red Hat Enterprise Linux or CentOS
...
+#. Upgrade the node as described in :ref:`upgrade_standalone`.
+   
+.. seealso::
+
+   |mongodb| Documentation: Upgrade a Replica Set
+       https://docs.mongodb.com/manual/release-notes/4.2-upgrade-replica-set/
+
+.. note::
+
+   Steps to upgrade from |mongodb-ce| with data encryption enabled to |PSMDB| are different. ``mongod`` requires an empty ``dbPath`` data directory because it cannot encrypt data files in place. It must receive data from other replica set members during the initial sync. Please refer to the :ref:`switch_storage_engines` for more information on migration of encrypted data. `Contact us <https://www.percona.com/about-percona/contact#us>`_ for working at the detailed migration steps, if further assistance is needed.
```
igorsol, line 158: Same is also true for situations when user wants to move from unencrypted storage to encrypted one or vice versa. Just my thoughts.

source/install/upgrade_from_mongodb.rst
akira-kurogane: Yes, "upgrade service" isn't used. "Upgrade procedure" is the most normal phrase I can think of. "Upgrade process" can be used too but procedure sounds better.

source/install/upgrade_from_mongodb.rst
akira-kurogane: "upgrade procedure" is best. "upgrade process" or "upgrade steps" would also be OK.

source/glossary.rst
```diff
@@ -39,3 +39,5 @@
      * aid developers in ensuring merge requests build and test on all platforms,
      * no known performance regressions (without a damn good explanation).
 
+  Rolling restart
+    A rolling restart (rolling upgrade) is shutting down and upgrading nodes one by one. The whole cluster remains operational and you achieve almost zero downtime during upgrade.
```
akira-kurogane:
```suggestion
    A rolling restart (rolling upgrade) is shutting down and upgrading nodes one by one. The whole cluster remains operational. There is no interruption to clients assuming the elections are short and all writes directed to the old primary use the retryWrite mechanism.
```

source/install/upgrade_from_mongodb.rst
```diff
@@ -100,5 +117,35 @@ Upgrading on Red Hat Enterprise Linux or CentOS
...
+1. Upgrade a secondary node in a replica set as described in :ref:`upgrade_standalone`. Use the instructions relevant to your operating system. 
+#. Wait for the node to rejoin with the replica set members and report the SECONDARY status. 
+#. Repeat the upgrade procedure on the remaining secondary nodes. 
+   
+Upgrade the primary node
+--------------------------------------------------------------------------------
+
+1. Step down the primary node: :command:`rs.stepDown()`.
+2. Wait for the remaining nodes to elect a new primary. Run :command:`rs.status()` to verify that the former primary node reports as SECONDARY.
+#. Upgrade the node as described in :ref:`upgrade_standalone`.
```
akira-kurogane:
```suggestion
Upgrade some but not all replica set nodes  
--------------------------------------------------------------------------------

1. Upgrade a node in a replica set as described in :ref:`upgrade_standalone`. Use the instructions relevant to your operating system. It is better to do it to the nodes that are currently secondary role so there won't be an extra election. If you do it to the node that is the current primary use :command:`rs.stepDown()` before shutting it down.
#. Wait for the node to rejoin with the replica set members, resync and report that it is in SECONDARY status. 
#. Optional: repeat the upgrade procedure on other but not all nodes.

Test the new version in the primary role before all nodes upgraded
----------------------------------------------------------------------------------------

This step is optional. Its purpose is to give a testing stage with the new version as primary whilst you still have at least one of the old version nodes. This will make rolling back a little quicker if you choose to do so.

1. Use :command:`rs.stepDown()` on the current primary to start an election and influence the primary role to switch to one of the new version nodes. If you have multiple nodes of the previous version use :command:`rs.freeze()` on them to make sure one of the new version nodes becomes primary.
 
Upgrade the last node(s)
--------------------------------------------------------------------------------

1. If any previous version nodes are the current primary node step them down: :command:`rs.stepDown()`.
2. Wait for the remaining nodes to elect a new primary. Run :command:`rs.status()` to verify that the former primary node reports as SECONDARY.
#. Upgrade the node as described in :ref:`upgrade_standalone`.
```

Hi Anastasia. I've added a middle section explaining the test-new-as-primary stage.

source/install/upgrade_from_mongodb.rst
akira-kurogane: > Same is also true for situations when user wants to move from unencrypted storage to encrypted one or vice versa.

I think they'll get the implication.

akira-kurogane (changes requested):
Hi Anastasia.

Please see suggestions in line.

source/install/upgrade_from_mongodb.rst
```diff
@@ -17,38 +18,46 @@ An in-place upgrade is done with existing data in the server.  Generally
...
+cases.  
 
 .. warning::
 
-   Before starting the upgrade process it is recommended to perform a full
+   Before starting the upgrade procedure, we recommend to perform a full
    backup of your data.
 
-The upgrade process depends on the distribution you are using:
+The sections below describe an in-place upgrade of a standalone instance and a replica set without encryption.
```
akira-kurogane: Encryption might be interpreted as network encryption.
```suggestion
The sections below describe an in-place upgrade of a standalone instance and a replica set. If using data-at-rest encryption please see extra note at end.
```

akira-kurogane (approved):
LGTM

sandraromanchenko (approved):

PR #288, opened 2020-10-21
PSMDB-675 Documented restore conditions

nastena1606:


sandraromanchenko (approved):

source/hot-backup.rst
```diff
@@ -37,6 +37,10 @@ with the error message, for example:
 
 .. rubric:: Restoring data from the backup
 
+.. note::
+
+   To guarantee consistency of oplog data, restore the backup into the node with the replication enabled (the ``replication.replSetname`` option is specified in the configuration file) as oplog is available in replica sets and clusters only. If you need a standalone node, restart this node with the replication disabled after the restore.
```
akira-kurogane:
```suggestion
   To guarantee consistency between the oplog and other collections you must restore the backup into a node with the replication enabled (i.e. the ``replication.replSetname`` option is specified in the configuration file).
```

I remove "Clusters" because they are besides the point. A hot backup is for one node's data only. One node can be a replica set by itself, but it can't be a cluster by itself.

I remove "if you need a standalone" part because if they want to do that, fine, they can just ignore the "local" db (it's only used for replication) and the oplog in it.

akira-kurogane (approved):
Hi Anastasia. Thanks for making this patch.

I ended up making some edits though sorry. No more to add after this though.

akira-kurogane:
@nastena1606 I've added new comment in PSMDB-675. Could you please take a look, and change the documentation from what my earlier request was.

https://jira.percona.com/browse/PSMDB-675?focusedCommentId=270825&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-270825

I guess the first commit here should be discarded completely and the advice changed to the idea of https://docs.mongodb.com/manual/tutorial/restore-replica-set-from-backup/#drop-the-local-database-if-it-exists-in-the-backup.

akira-kurogane (approved):
LGTM

PR #289, opened 2020-10-22
PSMDB-712-Removed connection pool limitation

nastena1606:
The connection pool supports was added, thus removed this liitation from the list for LDAP authorization


nastena1606:
Done in scope of https://github.com/percona/psmdb-docs/pull/305

PR #297, opened 2020-11-10
PSMDB-771 Reorganized sections in install instructions

nastena1606:
Moved Enable percona repository under the parent Install section


patrickbirch (approved):
In tip.run-all.root
Can you change the actions to gerunds to match the other docs?
Otherwise, LGTM.

PR #317, opened 2021-01-27
Psmdb 794 upgrade cluster instructions 4.4

nastena1606:


sandraromanchenko (approved):

akira-kurogane (commented):
Hi Anastasia.

Why should we duplicate the instructions on upgrading a cluster? The process is the same for Percona Server for MongoDB as it is for MongoDB Community (or MongoDB Enterprise).

nastena1606:
We have instructions for a single node and a replica set. I think it's logical to give at least brief instructions how to upgrade the cluster.

akira-kurogane:
I suggest we go the other way, that is delete the instructions for a single node and a replica set, and only point to the docs.mongodb.com documentation for that.

I wouldn't say the entire 'Upgrade from MongoDB Community' page should be deleted. It explains how the package repo is changed and shows the exact name of the package ("percona-server-mongodb"). Plus it mentions the fix needed in the mongod.conf regarding the pid file.

I feel too how making this decision is a harder job than just adding some more text content. But adding more content adds up to more maintenance work, plus more expectations that we keep on adding more and more content because readers would like to stay on one site. But we can't reproduce all that documentation.

source/install/upgrade_from_mongodb.rst
```diff
@@ -4,6 +4,11 @@
 Upgrading from |mongodb-ce|
 ================================================================================
 
+An in-place upgrade is done with existing data in the server.  Generally
+speaking, this is stopping the server, removing the old packages, installing the
+new server and starting it with the same data files.  While an in-place upgrade
+may not be suitable for high-complexity environments, it should work in most
+cases.
```
akira-kurogane:
```suggestion
An in-place upgrade is done with existing data in the server.  Generally
speaking, this is stopping the mongod service, removing the old packages, installing the
new server and starting it with the same db data directory.  While an in-place upgrade
may not be suitable for high-complexity environments, it should work in most
cases.
```

akira-kurogane: Regarding _"While an in-place upgrade_
_may not be suitable for high-complexity environments, it should work in most_
_cases."_

An in-place upgrade is suitable in all cases I think except the following:
* ephemeral hosts are used, ones that get new IP addresses and/or hostnames if they are restarted, and/or
* disk storage is ephemeral

The exceptions above are only situations you would get into with container-based systems or some cloud virtual machine services.

Classic on-prem servers, or VMs that are used like classic servers, can (and should i.m.o.) use in-place updates even if they are high-complexity in other ways.

akira-kurogane: So instead of "may not be suitable for high-complexity environments" I suggest "will not be suitable for environments that use ephemeral storage or host addresses"

source/install/upgrade_from_mongodb.rst
```diff
@@ -14,37 +19,34 @@ Upgrading from |mongodb-ce|
...
 
-   Before starting the upgrade procedure, we recommend to perform a full
-   backup of your data.
+Prerequisites
+==================
 
-The sections below describe an in-place upgrade of a standalone instance and a replica set. If using data-at-rest encryption please see extra note at end.
+Before you start the upgrade, update the |mongodb| configuration file
+(:file:`/etc/mongod.conf`) to contain the following settings.  Otherwise the
+|mongod| service will not be able to start.
```
akira-kurogane:
```suggestion
|mongod| service will not be able to start. Troubleshooting tip: If the "pidFilePath" setting in mongod.conf doesn't match the "PIDFile" option in the systemd "mongod" service unit the service will kill the mongod process after a timeout. 
```

akira-kurogane (changes requested):
Thanks for the reduction in content on this page Anastasia.

I have some edit suggestions for upgrade_from_mongodb.rst.

akira-kurogane (approved):
Thanks Anastasia. Looks good to me.

PR #328, opened 2021-02-11
PSMDB-812 Doc: Reworked ngram doc

nastena1606:
- Changed title
- Added content from github doc
- Restructured the doc

### DIFF
--- a/source/ngram-full-text-search.rst
+++ b/source/ngram-full-text-search.rst
@@ -1,18 +1,15 @@
 .. full-text-search.ngram:
 
 ================================================================================
-|ngram| Full Text Search
+Additional text search algorithm - *ngram* 
 ================================================================================
 
-The |ngram| full text search is useful for searching text for a specific string
-of characters in a field of a collection. To use |ngram|, create a text index on
-a collection setting the ``default_language`` parameter to **ngram**:
+The `ngram <https://en.wikipedia.org/wiki/N-gram>`_ text search algorithm is useful for searching text for a specific string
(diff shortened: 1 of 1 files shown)


source/ngram-full-text-search.rst
```diff
@@ -1,18 +1,15 @@
...
 
-The |ngram| full text search is useful for searching text for a specific string
-of characters in a field of a collection. To use |ngram|, create a text index on
-a collection setting the ``default_language`` parameter to **ngram**:
-
-.. code-block:: guess
-
-   mongo > db.collection.createIndex({name:"text"}, {default_language: "ngram"})
+The `ngram <https://en.wikipedia.org/wiki/N-gram>`_ text search algorithm is useful for searching text for a specific string
+of characters in a field of a collection. This feature is useful for such languages as Korean, Chinese, or Japanese. It
```
akira-kurogane:
```suggestion
of characters in a field of a collection. This feature can be used to find exact sub-string matches, which provides an alternative when parsing text from languages other than the list of European languages already supported by MongoDB Community's full text indexing. It
```
I wouldn't say that it is good for chinese, korean or japanese. Especially japanese has modifying endings on adjectives and verbs and ngram will not identify them as the same tokens.

At any rate I don't want to give the impression that CJK is well-supported by ngram.

akira-kurogane: I should have removed those claims about CJK support when I first saw that documentation; my bad.

akira-kurogane (changes requested):

PR #359, opened 2021-04-08
PSMDB-851 Added data-at-rest encryption to features list

nastena1606:
modified:   source/index.rst


nastena1606:
Akira, 3.6 includes MongoRocks as the storage engine. Are we leaving it for 3.6 or aligning the feature list with other versions?

akira-kurogane:
Leave it in 3.6. MongoRocks was deprecated but still present in 3.6;
removed in 4.0.

PR #379, opened 2021-06-02
PSMDB-878 Replaced SELinux config options with the node about setting…

nastena1606:
… up user policies for enforcing mode

modified: yum.rst


akira-kurogane (approved):
Thanks, that looks good to me!

PR #493, opened 2022-02-04
PSMDB-994 Enable Download PDF link in docs

nastena1606:
Amended the template to include the Download PDF link
renamed:    source/_static/_templates/sidebar.html -> source/_static/_templates/theme/sidebar.html


LenzGr (approved):
LGTM! Thank you.

PR #495, opened 2022-02-16
PSMDB-1003 Updated SASL auth page

nastena1606:
Removed misleading note about user roles

modified:   source/sasl-auth.rst

### DIFF
--- a/source/sasl-auth.rst
+++ b/source/sasl-auth.rst
@@ -207,6 +207,8 @@ The configuration consists of the following steps:
 * Creating a user with the **root** privileges. This user is required to log in to |PSMDB| after the external authentication is enabled.
 * Editing the configuration file to enable the external authentication
 
+.. _root-user:
+
 Create a root user
 **********************************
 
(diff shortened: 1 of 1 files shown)


igorsol (approved):

source/sasl-auth.rst
```diff
@@ -238,37 +240,43 @@ Restart the ``mongod`` service:
 
    $ sudo systemctl restart mongod
 
-When everything is configured properly, you can use the :ref:`commands`.
+Add an external user to |PSMDB|
+*************************************
 
-.. _commands:
+User authentication is done by mapping a user object against a user created in the ``$external`` database. Thus, at this step, you create the user in the ``$external`` database and assign the roles and privileges to them. Note that the username must exactly match the name of the user object on the LDAP server.
```
igorsol: At first I was confused by "mapping a user object against a user created in the $external database". I could not understand what user object we are talking about. After some thinkkuing I think this is "LDAP user object". So I think whole phrase should be "mapping a LDAP user object against a user created in the $external database"

igorsol (changes requested):

igorsol (approved):

PR #534, opened 2022-05-10
PSMDB-1050 Added a missing code sample to inmemory doc

nastena1606:
modified:   source/inmemory.rst

### DIFF
--- a/source/inmemory.rst
+++ b/source/inmemory.rst
@@ -24,6 +24,13 @@ storage engine using the ``--storageEngine`` command-line option when you start
 ``mongod``.  Alternatively, you can set the ``storage.engine`` variable in the
 configuration file (by default, :file:`/etc/mongod.conf`):
 
+.. code-block:: yaml
+
+   storage:
+     dbPath: <dataDir>
+     engine: inMemory
+
+
 .. seealso::
 
    |mongodb| Documentation: Configuration File Options


sandraromanchenko (approved):

source/inmemory.rst
```diff
@@ -24,6 +24,13 @@ storage engine using the ``--storageEngine`` command-line option when you start
 ``mongod``.  Alternatively, you can set the ``storage.engine`` variable in the
 configuration file (by default, :file:`/etc/mongod.conf`):
 
+.. code-block:: yaml
+
+   storage:
+     dbPath: <backupDir>
```
igorsol: dbPath is not for "backupDir". Moreover dbPath is not necessary when using inMemory storage engine.

sandraromanchenko: @igorsol it's not mandatory - true (default /data/db will be used instead), but it's still required - directory is used for at least diagnostic.data and storage.bson 
ls -lah /var/lib/mongodb/
drwx------  2 root   root   4.0K May 10 08:30 diagnostic.data
-rw-------  1 root   root      5 May 10 08:29 mongod.lock
-rw-------  1 root   root     93 May 10 08:29 storage.bson

igorsol (approved):
thanks!

PR #564, opened 2022-06-10
PSMDB-1076 Updated the features comparison table

nastena1606:
modified:   source/comparison.rst

### DIFF
--- a/source/comparison.rst
+++ b/source/comparison.rst
@@ -21,29 +21,27 @@ in MongoDB Enterprise Edition.
      - * WiredTiger_ (default)
        * In-Memory_ (Enterprise only)
    * - Encryption-at-Rest
-     - * Key server = Hashicorp Vault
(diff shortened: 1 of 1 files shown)


LenzGr:
Looks good, thank you! Looking at the [resulting table](https://psmdb-docs-5-0-pr-564.onrender.com/comparison.html#comparison--page-root), I wonder if you could make a few additional improvements?

Currently, there are links to some features in the left column (e.g. "Hot Backup" and "Audit Logging"), while most other features don't include links to the corresponding PSMDB documentation. Would it be possible to be more consistent here? Instead of adding more links to the first column (converting the feature names into links), I'd actually suggest using the "YES" words in the second (PSMDB) column for linking to the respective documentation related to that feature. What do you think?

nastena1606:
@LenzGr corrected per your comments

PR #589, opened 2022-08-25
PSMDB-1046 Updated kmipKeyIdentifier option description

nastena1606:


source/kmip.rst
```diff
@@ -37,7 +41,7 @@ Starting with version 4.4.15-15, you can specify multiple KMIP servers for failo
      - The path to the PEM file with the KMIP client private key and the certificate chain. The database server uses this PEM file to authenticate the KMIP server.
    * - ``--kmipKeyIdentifier``
      - string
-     - Mandatory. The name of the KMIP key. If the key does not exist, the database server creates a key on the KMIP server with the specified identifier.
+     - Optional starting with version 4.4.16-16. The identifier of the KMIP key. If the key does not exist, the database server creates a key on the KMIP server with the specified identifier. When you specify the identifier, the key with such a name must exist on the key storage. You can only use this setting for the first time you enable encryption.
```
ktrushin: I'd replace  `the key with such a name must exist` with `the key with such an ID must exist`. In some KMIP-servers (e.g. PyKMIP) a key has the associated `name` property that is not unique among all the keys. Using "name" in the description may confuse the user into specifying a key name rather than a key identifier.

This is the second instance of the `the key with such a name must exist` wording. Please don't also forget to change the first one above.

ktrushin (changes requested):

source/kmip.rst
nastena1606: Updated. Thank you for the explanation

ktrushin (approved):

PR #603, opened 2022-10-26
PSMDB-1152 Described backupCursor and backupCursorExtend

nastena1606:


source/backup-cursor.rst
```diff
@@ -0,0 +1,95 @@
+.. _backup-cursor:
+
+============================================================
+$backupCursor and $backupCursorExtended aggregation stages
+============================================================
+
+``$backupCursor`` and ``$backupCursorExtended`` aggregation stages allow developing your custom backup solution: 
+
+* ``$backupCursor`` outputs the list of files and their size to copy.  
+* ``$backupCursorExtended`` outputs the list of WiredTiger transaction log files that have been updated or newly added since the ``$backupCursor`` was first run. Saving these files enables restoring the database to any arbitrary time between the ``$backupCursor`` and ``$backupCursorExtend`` execution times.
```
igorsol: replace $backupCursorExtended with $backupCursorExtend everywhere

source/backup-cursor.rst
```diff
@@ -0,0 +1,95 @@
...
+
+   bkCsr
+   [
+     {
+       metadata: {
+         backupId: UUID("35c34101-0107-44cf-bdec-fad285e07534"),
+         dbpath: '/var/lib/mongodb',
+         oplogStart: { ts: Timestamp({ t: 1666631297, i: 1 }), t: Long("-1") },
+         oplogEnd: { ts: Timestamp({ t: 1666631408, i: 1 }), t: Long("1") },
+         checkpointTimestamp: Timestamp({ t: 1666631348, i: 1 })
```
igorsol, line 45: We need to warn readers that `oplogStart`, `oplogEnd`, `checkpointTimesatmp` values may be absent if $backupCursor is executed on standalone node.

source/backup-cursor.rst
```diff
@@ -0,0 +1,95 @@
...
+       fileSize: Long("104857600")
+     },
+     {
+       filename: '/var/lib/mongodb/WiredTigerHS.wt',
+       fileSize: Long("20480")
+     },
+     ....
+
+Store the ``metadata`` document somewhere, because you need to pass the ``backupId`` parameter from this document as the input parameter for the ``$backupCursorExtended`` stage. Also you need the ``oplogEnd`` timestamp.
+Make sure that the ``$backupCursor`` is complete on all shards in your cluster.
```
sandraromanchenko: This seems to me a bit confusing - can we add more detailed description like if user needs backup of single node in RS then he/she needs to run both stages and save the result, if intent is backup of the cluster then the same procedure should run on each shard? Now it might seem that this should work only if there is cluster.

source/backup-cursor.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Run both aggregation stages on one node from each replica set. You can use either the primary or the secondary node. Note that since the secondary node may lag in syncing the data from the primary one, you will have to wait for the exact same time before running the ``$backupCursorExtended``.
+
+Get a list of all files to copy with $backupCursor
+---------------------------------------------------------------
+
+.. code-block:: javascript
+
+    var bkCsr = db.getSiblingDB("admin").aggregate([{$backupCursor: {}}])
+    var bkCsrMetadata = bkCsr.next().metadata
```
sandraromanchenko: maybe it's worth removing var in second command so that user actually gets some output, since when cursor returned to a variable using var, it doesn't automatically iterate and it might be confusing)

var bkcsr = db.getSiblingDB("admin").aggregate([{$backupCursor: {}}])
bkCsrMetadata = bkcsr.next().metadata
{
  backupId: new UUID("2fc8ac28-0fbe-4d42-962c-bd1d236f3394"),
  dbpath: '/home/vagrant/percona-server-mongodb-6.0.2-1-x86_64.glibc2.17/nodes/node1/db',
  oplogStart: { ts: Timestamp({ t: 1666641030, i: 1 }), t: Long("-1") },
  oplogEnd: { ts: Timestamp({ t: 1666944139, i: 2 }), t: Long("6") },
  checkpointTimestamp: Timestamp({ t: 1666944086, i: 3 })
}

source/backup-cursor.rst
```diff
@@ -0,0 +1,95 @@
+.. _backup-cursor:
+
+============================================================
+$backupCursor and $backupCursorExtended aggregation stages
+============================================================
+
+``$backupCursor`` and ``$backupCursorExtended`` aggregation stages allow developing your custom backup solution:
```
sandraromanchenko: it's valid use-case but IMHO not the feature description) Maybe add smth like - New aggregation pipeline stages expose WiredTiger API which gives a way to make consistent backup by listing all files that should be copied for backup and freezing those files so you can copy them without them being deleted or necessary parts within them being overwritten - just compilation of Akira's thoughts:)

sandraromanchenko (changes requested):

source/backup-cursor.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Store the ``metadata`` document somewhere, because you need to pass the ``backupId`` parameter from this document as the input parameter for the ``$backupCursorExtended`` stage. Also you need the ``oplogEnd`` timestamp.
+Make sure that the ``$backupCursor`` is complete on all shards in your cluster. 
+
+Run ``$backupCursorExtended`` to retrieve the WiredTiger transaction logs
+--------------------------------------------------------------------------
+
+.. code-block:: javascript
+
+   var bkExtCsr = db.aggregate([{$backupCursorExtend: {backupId: bkCsrMetadata.backupId, timestamp: new Timestamp(1666631418, 1)}}])
```
sandraromanchenko, line 64: Same here - remove var so that user gets output which is cool for educational purposes, IMHO)

bkExtCsr = db.aggregate([{$backupCursorExtend: {backupId: bkCsrMetadata.backupId, timestamp: new Timestamp(1666944242, 1)}}])
[
  {
    filename: '/home/vagrant/percona-server-mongodb-6.0.2-1-x86_64.glibc2.17/nodes/node1/db/journal/WiredTigerLog.0000000011'
  },
  {
    filename: '/home/vagrant/percona-server-mongodb-6.0.2-1-x86_64.glibc2.17/nodes/node1/db/key.db/WiredTigerLog.0000000009'
  },
  {
    filename: '/home/vagrant/percona-server-mongodb-6.0.2-1-x86_64.glibc2.17/nodes/node1/db/key.db/WiredTigerLog.0000000010'
  },
  {
    filename: '/home/vagrant/percona-server-mongodb-6.0.2-1-x86_64.glibc2.17/nodes/node1/db/key.db/WiredTigerLog.0000000007'
  },
  {
    filename: '/home/vagrant/percona-server-mongodb-6.0.2-1-x86_64.glibc2.17/nodes/node1/db/key.db/WiredTigerLog.0000000008'
  },
  {
    filename: '/home/vagrant/percona-server-mongodb-6.0.2-1-x86_64.glibc2.17/nodes/node1/db/key.db/WiredTigerLog.0000000011'
  }
]

source/backup-cursor.rst
igorsol: but if cursor is auto iterated then user cannot iterate it to do actual backup because cursor is already exhausted

source/backup-cursor.rst
igorsol: I like this suggestion

source/backup-cursor.rst
nastena1606: How to define the timestamp for backupCursorExtend in this case?

source/backup-cursor.rst
sandraromanchenko: as decided we can remove VAR variable here
bkCsrMetadata = bkcsr.next().metadata

source/backup-cursor.rst
```diff
@@ -0,0 +1,96 @@
...
+   Note that when running ``$backupCursor`` in a single node deployment, ``oplogStart``, ``oplogEnd``, ``checkpointTimesatmp`` values may be absent. This is because single node deployments don't have oplogs.   
+
+Run ``$backupCursorExtend`` to retrieve the WiredTiger transaction logs
+--------------------------------------------------------------------------
+
+Pass the ``backupId`` from the metadata document as the first parameter. For the ``timestamp`` parameter, use the maximum (latest) value among the ``oplogEnd`` timestamps from all shards and config server replica set. This will be the target time to restore.
+
+.. code-block:: javascript
+
+   bkExtCsr = db.aggregate([{$backupCursorExtend: {backupId: bkCsrMetadata.backupId, timestamp: new Timestamp(1666631418, 1)}}])
```
sandraromanchenko: but we need to return VAR here - my bad, made a comment last minute in a hurry) 

var bkExtCsr = db.aggregate([{$backupCursorExtend: {backupId: bkCsrMetadata.backupId, timestamp: new Timestamp(1666631418, 1)}}])

source/backup-cursor.rst
```diff
@@ -0,0 +1,96 @@
+.. _backup-cursor:
+
+============================================================
+$backupCursor and $backupCursorExtend aggregation stages
+============================================================
+
+``$backupCursor`` and ``$backupCursorExtend`` aggregation stages expose the WiredTiger API which allows making consistent backups. Using these stages enables you to list the files to copy and freeze those files so you can copy them without them being deleted or necessary parts within them being overwritten.
```
sandraromanchenko: My new suggestion ^_^
Using these stages allows listing and freezing database files so you can copy them without possibility for files being deleted or necessary parts within them being overwritten.

source/backup-cursor.rst
igorsol: agree

source/backup-cursor.rst
sandraromanchenko: backupCursorExtend won't be available at all for PSMDB in standalone mode if oplog is disabled

source/backup-cursor.rst
nastena1606: updated

source/backup-cursor.rst
nastena1606: Updated

source/backup-cursor.rst
```diff
@@ -0,0 +1,97 @@
...
+They are available in |PSMDB| starting with version 4.4.6-8. 
+
+Percona provides `Percona Backup for MongoDB (PBM) <https://www.percona.com/doc/percona-backup-mongodb/index.html>`_ – a light-weight open source solution for consistent backups and restores across sharded clusters. PBM relies on these aggregation stages for physical backups and restores. However, if you wish to develop your own backup application, this document describes the ``$backupCursor`` and ``$backupCursorExtend`` aggregation stages.
+
+Usage 
+=============
+
+You can run these stages in any type of MongoDB deployment. If you need to a back up a single node in a replica set, first run the ``$backupCursor``, then the ``$backupCursorExtend`` and save the output files to the backup storage.
+
+To make a consistent backup  of the whole replica set / a sharded cluster, run both aggregation stages on one node from each replica set. It can be either the primary or the secondary node. Note that since the secondary node may lag in syncing the data from the primary one, you will have to wait for the exact same time before running the ``$backupCursorExtend``.
```
sandraromanchenko: Please change to 
To make a consistent backup of the sharded cluster, run both aggregation stages on one node from each replica set.

sandraromanchenko (approved):

janwieremjewicz (approved):

igorsol (approved):

PR #666, opened 2023-02-02
Fix version in comparison page

sleto-it:
All the other branches are OK. 
The issue only occurs on the 5.0 branch.


nastena1606 (approved):
LGTM Thank you for spotting it!

PR #671, opened 2023-03-07
PSMDB-1148 Updated vault description

nastena1606:
modified:   docs/vault.md

### DIFF
--- a/docs/vault.md
+++ b/docs/vault.md
@@ -17,6 +17,11 @@ with versioning enabled.
 | vaultServerName      | security.vault.serverName | string | The IP address of the Vault server|
 | vaultPort            | security.vault.port       | int    | The port on the Vault server|
 | vaultTokenFile       | security.vault.tokenFile  | string | The path to the vault token file. The token file is used by MongoDB to access HashiCorp Vault. The vault token file consists of the raw vault token and does not include any additional strings or parameters. <br> <br> Example of a vault token file: <br> <br> `s.uTrHtzsZnEE7KyHeA797CkWA`|
+| vaultSecret          | security.vault.secret     | string | The path to the Vault secret. The Vault secret path format must be ```<secrets_engine_mount_path>/data/<custom_path>``` <br> <br> where: <br> - ``<secrets_engine_mount_path>`` is the path to the Key/Value Secrets Engine v2; <br> - ``data`` is the mandatory path prefix required by Version 2 API; <br> - ``<custom_path>`` is the path to the specific secret. <br> <br> Example: `secret_v2/data/psmdb-test/rs1-27017` <br><br> Starting with version 4.4.19-19, a distinct Vault secret path for every replica set member is no longer mandatory. In earlier versions, it is recommended to use different secret paths for every database node in the entire deployment to avoid issues during the master key rotation.|
+| vaultSecretVersion | security.vault.<br>secretVersion | unsigned long | (Optional) The version of the Vault secret to use | 
+| vaultRotateMasterKey | security.vault.<br>rotateMasterKey| switch | When enabled, rotates the master key and exits |
+| vaultServerCAFile    | security.vault.<br>serverCAFile | string | The path to the TLS certificate file |
+| vaultDisableTLSForTesting | security.vault.<br>disableTLSForTesting | switch | Disables secure connection to Vault using SSL/TLS client certificates|
 
 **Config file example**
 
@@ -65,7 +70,7 @@ You have the following options of how to target a particular namespace when conf
 
     * [Secure Multi-Tenancy with Namespaces](https://learn.hashicorp.com/tutorials/vault/namespaces)
 
-## Key Rotation
+## Key rotation
 
 Key rotation is replacing the old master key with a new one. This process helps to comply with regulatory requirements.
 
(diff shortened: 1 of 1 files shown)


docs/vault.md
```diff
@@ -17,6 +17,10 @@ with versioning enabled.
 | vaultServerName      | security.vault.serverName | string | The IP address of the Vault server|
 | vaultPort            | security.vault.port       | int    | The port on the Vault server|
 | vaultTokenFile       | security.vault.tokenFile  | string | The path to the vault token file. The token file is used by MongoDB to access HashiCorp Vault. The vault token file consists of the raw vault token and does not include any additional strings or parameters. <br> <br> Example of a vault token file: <br> <br> `s.uTrHtzsZnEE7KyHeA797CkWA`|
+| vaultSecret          | security.vault.secret     | string | The path to the vault secret. The vault secrets path format must be ```<vault_secret_mount>/data/<custom_path>``` <br> <br> where: <br> - ``<vault_secret_mount>`` is your Vault KV Secrets Engine; <br> - ``data`` is the mandatory path prefix required by Version 2 API; <br> - ``<custom_path>`` is your secrets path. <br> <br> Example: `secret_v2/data/psmdb-test/rs1-27017` <br> Starting with version 4.4.19-19, a distinct vault secret for every replica set member is no longer mandatory. In earlier versions, it is recommended to use different secret paths for every database node.|
```
ktrushin: I'd like to suggest a couple of modifications to the description.
- Starting the word "Vault" with an upper-case letter in general writing (not in option or variable names). For instance,  `to the vault secret` should become `to the Vault secret`. This is how the authors name their sofware product. For instance, please see the phrase "This documentation covers the main concepts of Vault" in the [docs](https://developer.hashicorp.com/vault/docs)
- To my best knowledge, the phrase `secrets path` is not absolutely correct, if you talk about a path to a single secret. When a noun is used in the role of an adjective, it is usually in a singular form, e.g. "project manager". Though a person with this job can govern multiple projects simultaneously, we _don't_ say "project**s** manager". When it comes to the word "secret", things become more complex because it can be both a [noun](https://www.oxfordlearnersdictionaries.com/definition/english/secret_2) and an [adjective](https://www.oxfordlearnersdictionaries.com/definition/english/secret_1). But I'd use `secret path` (without `s` at the end) anyway. Please correct me if I am wrong. I would still use the phrase "secret**s** engine" (with `s` at the end of "secrets") though, for better consistency with Vault [documentation](https://developer.hashicorp.com/vault/tutorials/secrets-management/versioned-kv).
- `vault_secret_mount` seems to be quite a vague term. I'd replace it with `secrets_engine_mount_path`
- To remove even a tiny possibility of ambiguity, I'd use `Key/Value Secrets Engine v2` instead of `KV Secrets Engine` in the phrase `your Vault KV Secrets Engine`. The former is straight from the Vault's [docs](https://developer.hashicorp.com/vault/tutorials/secrets-management/compare-kv-versions?in=vault%2Fsecrets-management). On top of that, `vault_secret_mount` is not the engine itself, it is a path to it. I would introduce the word "path" to the phrase. Finally, I don't think the word "your" is appropriate in the formal technical text. To sum up, I'd rewrite the whole phrase as  `the path to the Key/Value Secrets Engine v2`.
- Transforming `is your secrets path` to the  `is the path to the specific secret`
- Refining the phrase "distinct vault secret for every replica set member is no longer mandatory" to "distinct Vault secret *path* for each database node in the deployment is no longer mandatory". In the latter phrase, it is emphasized that the new version lifts the restriction on paths rather than the keys themselves. Moreover, the fixed issue would have arisen even if two nodes from different replica sets had the same secret key path, thus, it is better to speak about the whole deployment.
- The phrase "recommended to use different secret paths for every database node" kind of logically follows from the previous sentence and there is no need to tell the same information twice. Instead, I'd say that "In earlier versions, common Vault secret path on two or more nodes could result in issues during master key rotation"

ktrushin: Right after the `vaultSecret` option, I'd add the description of the option `vaultSecretVersion` as follows:
```
| vaultSecretVersion | security.vault.secretVersion | unsigned long long | The version of the Vault secret to use (optional)
```

docs/vault.md
```diff
@@ -85,7 +89,11 @@ Rotating the master key process also re-encrypts the keystore using the new mast
 
 ### Key rotation in replica sets
 
-Every `mongod` node in a replica set must have its own master key. The key rotation steps are the following:
+Starting with version 4.4.19-19, you can use the same master key for all replica set members. Vault assigns different versions to the master key stored at the same path. The `mongod` server identifies the master key by its version and the path and uses that identifier to read the correct master key during the restart.
```
ktrushin:
```suggestion
Starting with version 4.4.19-19, the same secret path for all the nodes in the deployment can be used. Vault assigns different versions to the secrets stored at the same path. The `mongod` uses the `{<secret_path>, <secret_version>}` pair as an identifier of the master key and remembers that identifier internally so that the correct master key is retrieved from the Vault server during the restart.
```

docs/vault.md
```diff
@@ -85,7 +89,11 @@ Rotating the master key process also re-encrypts the keystore using the new mast
 
 ### Key rotation in replica sets
 
-Every `mongod` node in a replica set must have its own master key. The key rotation steps are the following:
+Starting with version 4.4.19-19, you can use the same master key for all replica set members. Vault assigns different versions to the master key stored at the same path. The `mongod` server identifies the master key by its version and the path and uses that identifier to read the correct master key during the restart.  
+
+In versions 4.4.18-18 and earlier, every `mongod` node in a replica set must have its own master key.
```
ktrushin:
```suggestion
In versions 4.4.18-18 and earlier, each `mongod` node in the deployment must have a distinct path to the master key on a Vault server
```

ktrushin (changes requested):

sandraromanchenko (approved):

ktrushin (approved):

PR #674, opened 2023-03-13
V2 GitHub Actions are deprecated, bump to v3

deejay1:
As per message - v2 actions are deprecated and they should be bumped to v3


nastena1606:
thanks. updated in scope of https://github.com/percona/psmdb-docs/pull/715

PR #754, opened 2023-08-21
PSMDB-1286 Release notes 7.0

nastena1606:


docs/release_notes/7.0.2-1.md
```diff
@@ -0,0 +1,46 @@
+# Percona Server for MongoDB 7.0.2-1 (Release Candidate) (2023-09-28) 
+
+[Installation](../install/index.md){.md-button}
+
+We are pleased to announce the release of the new major version of Percona Server for MongoDB 7.0.2-1 – the drop in replacement for MongoDB Community Edition. Percona Server for MongoDB supports MongoDB community protocols and drivers and enhances it with the enterprise-grade features for free, enabling you to effectively meet the requirements of your organization in terms of high-availability, reliability and data security. 
+
+## Upgrade considerations
+
+Starting with version 7.0, journaling is enabled by default in Percona Server for MongoDB and the `storage.journal.enabled` configuration option has no effect. Before the upgrade, edit the configuration file and remove the `storage.journal.enabled` option to ensure successful database start after the upgrade. See the [upgrade guide](../install/upgrade-from-60.md) for exact procedure.
```
igorsol: From this text user might think that `storage.journal.enabled` will prevent norml server start which is not true.
In fact we implemented special handling for the `storage.journal.enabled` option. In PSMDB 7.0 this option will be ignored and special warning explaining the option is ignored will be logged. See PSMDB-1335.

variables.yml
```diff
@@ -1,6 +1,6 @@
 # PBM Variables set for HTML output
 # See also mkdocs.yml plugins.with-pdf.cover_subtitle and output_path
 
-release: '6.0.8-6'
-version: '6.0'
-release_date: 2023-08-08
+release: '7.0.2-1'
+version: '7.0'
+
```
igorsol, line 6: No "release_date: 2023-09-28" here?

igorsol (changes requested):

variables.yml
nastena1606: It's not used in the docs so removed as an extra source of error

docs/release_notes/7.0.2-1.md
nastena1606: Why do we ask users to remove this option from config file before the upgrade if this option is ignored?

nastena1606: Reworked

igorsol (approved):

docs/release_notes/7.0.2-1.md
```diff
@@ -0,0 +1,46 @@
...
+* The [AutoMerger](https://www.mongodb.com/docs/v7.0/core/automerger-concept/#std-label-automerger-concept) now merges contiguous chunks that reside on the same shard. This reduces the sharding table, avoids unnecessary chunk migrations and improves overall performance. The AutoMerger is enabled by default.
+* The `findOne`, `updateOne` and `deleteOne` command behavior is now aligned for both sharded and unsharded collections. This allows you to run these commands without having to target a specific shard. This comes handy when you need to update documents on multiple shards and you don’t know exactly on what shard the data resides.
+
+### Security
+
+* Improve security and safety of your sensitive data with [queryable encryption](https://www.mongodb.com/docs/v7.0/core/queryable-encryption/#std-label-qe-manual-feature-qe). This mechanism enables you to encrypt/decrpyt the data on the client side, store it as encrypted on the database and run queries against the encrypted data. This way your data remains encrypted and secure throughout its whole lifecycle. The community addition supports explicit encryption only.
```
janwieremjewicz: Community edition rather than addition

janwieremjewicz (approved):
I am missing information in our release notes about no downgrade possibility that is an exact outcome of Upstream changes

sandraromanchenko (approved):

PR #882, opened 2024-07-08
PSMDB-1476 added info what tarball suits which OS

nastena1606:
Explained tarball types

modified:   docs/install/tarball.md

### DIFF
--- a/docs/install/tarball.md
+++ b/docs/install/tarball.md
@@ -8,6 +8,44 @@ There are two tarballs available:
 
 * `percona-server-mongodb-{{release}}-x86_64.glibc2.35.tar.gz` is the tarball for Ubuntu 22.04.
 
+To check which `glibc` version your system is using, run the following command:
+
+```{.bash data-prompt="$"}
+$ ldd --version
+```
+
+??? example "Sample output"
+
+    ```{.text .no-copy}
+    ldd (GNU libc) 2.28
+    Copyright (C) 2018 Free Software Foundation, Inc.
+    This is free software; see the source for copying conditions.  There is NO
+    warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    Written by Roland McGrath and Ulrich Drepper.
+    ```
+
+### What tarballs to choose
+
+The following table helps you understand what tarball to download based on the glibc version of your operating system.
+
+| Operating system | Tarball to download | glibc version  |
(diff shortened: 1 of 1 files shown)


sandraromanchenko (approved):

docs/install/tarball.md
```diff
@@ -8,6 +8,44 @@ There are two tarballs available:
...
+
+### What tarballs to choose
+
+The following table helps you understand what tarball to download based on the glibc version of your operating system.
+
+| Operating system | Tarball to download | glibc version  |
+|------------------|---------------------|----------------|
+| Ubuntu 22.04         | percona-server-mongodb-{{release}}-x86_64.glibc2.35.tar.gz     | glibc2.35 |
+| Ubuntu 20.04         | percona-server-mongodb-{{release}}-x86_64.glibc2.17.tar.gz     | glibc2.31 |
+| Debian 10            | percona-server-mongodb-{{release}}-x86_64.glibc2.17.tar.gz     | glibc2.28 |
```
sleto-it: I would suggest is to invert the Debian 10 and 11 rows, so things are in descending order:

- Debian 11
- Debian 10

We have descending order for Ubuntu and RHEL, good to keep it consistent.  Same on the MySQL page, it is descending: https://docs.percona.com/percona-server/innovation-release/binary-tarball-names.html#binary-tarball-file-name-organization

docs/install/tarball.md
```diff
@@ -8,6 +8,44 @@ There are two tarballs available:
...
+
+    ```{.text .no-copy}
+    ldd (GNU libc) 2.28
+    Copyright (C) 2018 Free Software Foundation, Inc.
+    This is free software; see the source for copying conditions.  There is NO
+    warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    Written by Roland McGrath and Ulrich Drepper.
+    ```
+
+### What tarballs to choose
```
sleto-it: Should probably be headline 2? We are going from 1 to 3, and then go back to 2 for "Precondition".  Check also "Tarball types".

sleto-it (approved):
Shared some minor comments.

docs/install/tarball.md
nastena1606: Fixed

docs/install/tarball.md
nastena1606: Updated

PR #893, opened 2024-07-23
Fixed rendering of example

nastena1606:
docs/install/upgrade-from-mongodb.md


alina-derkach-oaza:
The example in  "Upgrade on Red Hat Enterprise Linux and derivatives" also does not work.

nastena1606:
Thanks, fixed!

alina-derkach-oaza (approved):

PR #896, opened 2024-08-01
PSMDB-1283 Added KMIP key state polling

nastena1606:
modified:   docs/kmip.md


docs/kmip.md
```diff
@@ -73,11 +73,33 @@ Starting with version 5.0.11-10, the `kmipKeyIdentifier` option is no longer man
 | **Type**            | int|
 | **Description**     | The time to wait for the response from the KMIP server. Min value: 1000. Default: 5000. <br><br>If the `connectRetries` setting is specified, the `mongod` waits up to the value specified with `connectTimeoutMS` for each retry.|
 
-# Key rotation
+| Configuration file  | {{optionlink('security.kmip.activateKeys')}}|
+|-------------------- | --------------------|
+| **Command line**    | `kmipActivateKeys`|
+| **Type**            | boolean|
+| **Description**     | Controls the transition of the master key from the Pre-Active to the Active state. Enabled by default. Available starting with version 5.0.28-24.|
```
ktrushin: From my perspective, the phrase "Controls the transition" is too vague. It can imply that transition is a tunable process with configurable parameters and `kmipActivateKeys` is one of them.
I suggest rewriting to
`Determines whether a newly created master encryption key is transitioned to the Active state`.

docs/kmip.md
```diff
@@ -73,11 +73,33 @@ Starting with version 5.0.11-10, the `kmipKeyIdentifier` option is no longer man
...
+
+| Configuration file  | {{optionlink('security.kmip.keyStatePollingSeconds')}}|
+|-------------------- | --------------------|
+| **Command line**    | `kmipKeyStatePollingSeconds`|
+| **Type**            | int|
+| **Description**     | Sets the period in seconds to check the state of the master encryption key. Default: 900. If the master key for a node is in the state other than Pre-Active or Active, the node logs the error and shuts down. Available starting with version 5.0.28-24.|
```
ktrushin: Pre-Active is a tricky topic in older major versions. I would not mention it here at all and elaborate later.

docs/kmip.md
```diff
@@ -73,11 +73,33 @@ Starting with version 5.0.11-10, the `kmipKeyIdentifier` option is no longer man
...
+
+
+## Key rotation
+
+Starting with version 5.0.8-7, the support for [master key rotation](https://www.mongodb.com/docs/manual/tutorial/rotate-encryption-key/#kmip-master-key-rotation) is added. This enables users to comply with data security regulations when using KMIP.
+
+## Key state polling
+
+When Percona Server for MongoDB generates a new master encryption key for every node, it automatically registers the key on the KMIP server with a `Pre-Active` state. Starting with version 5.0.28-24, Percona Server for MongoDB automatically activates the master encryption key and checks (polls) its state within a defined period. If a master encryption key for a node is not in the `Active` state, the node reports an error and shuts down. This process helps security engineers identify the nodes that require out-of-schedule master key rotation.
```
ktrushin: I think the phrase `When Percona Server for MongoDB generates a new master encryption key for every node` sounds somewhat strange. As if there is a dedicated entity named "Percona Server for MongoDB" and a group of other similar entities called "nodes". The phrase reads as the first entity generated all the keys and hands them out to the "nodes", one key per node.
Please consider replacing the phrase with `When an instance of Percona Server for MongoDB generates a new master encryption, ...`

docs/kmip.md
```diff
@@ -73,11 +73,33 @@ Starting with version 5.0.11-10, the `kmipKeyIdentifier` option is no longer man
...
+
+The `security.kmip.activateKeys` configuration file option controls the transition of the master encryption key to the `Active` state as the startup. This option is enabled by default and is mandatory for the server to start starting with version 8.0.0. The `security.kmip.keyStatePollingSeconds` configuration file option sets the period, in seconds, to poll the master key state.
```
ktrushin: I see several issues in this paragraph:
1. As above, the phrase ``controls the transition of the master encryption key to the `Active` state`` can be misleading. I'd replace it with ``determines whether a new master encryption key is transitioned to the `Active` state``.
2. The phrase `as the startup` should be probably replaced with `at the startup`. Do we need the article `the` here?
3.  I would completely eliminate the phase `and is mandatory for the server to start starting with version 8.0.0`. It  conveys "the option is mandatory" as if one needed to always manually enable option starting from 8.0.0, which is clearly not true. It is an active key that is mandatory since 8.0.0, but we'll will discuss that in another comment.

docs/kmip.md
```diff
@@ -73,11 +73,33 @@ Starting with version 5.0.11-10, the `kmipKeyIdentifier` option is no longer man
...
+
+The `security.kmip.activateKeys` configuration file option controls the transition of the master encryption key to the `Active` state as the startup. This option is enabled by default and is mandatory for the server to start starting with version 8.0.0. The `security.kmip.keyStatePollingSeconds` configuration file option sets the period, in seconds, to poll the master key state. 
+
+The master key state polling functionality is particularly useful in cluster deployments with hundreds of nodes. If some master keys are compromised, security engineers change their state from Active so that the nodes encrypted with these keys identify themselves. This approach allows the security engineers to rotate master keys only on the affected nodes instead of the entire cluster, thus reducing maintenance time and administrative efforts while improving overall cluster health and security.
```
ktrushin: I'd replace the `reducing maintenance time and administrative efforts while improving overall cluster health and security` phrase with the one suggested by Radek: `reducing the mean time to resolve (MTTR) compromised encryption key incidents`

docs/kmip.md
```diff
@@ -73,11 +73,33 @@ Starting with version 5.0.11-10, the `kmipKeyIdentifier` option is no longer man
...
```
ktrushin, line 126: Here is the best place to describe what happens to `Pre-Active`.

```suggestion
All 5.0.* versions starting 5.0.28 tolerate already existing `Pre-Active` keys. If at startup Percona Server for MongoDB detects that data directory is encrypted with an exisging key in the `Pre-Active` state, it logs a warning and continues as usual. In that case, Percona Server for MongoDB does not do periodic key state polling whatever the value of `kmipKeyStatePollingSeconds` is. An active master encryption key is mandatory since version 8.0.0. To get an active key, one can do a master key rotation or active a key manually.
```

ktrushin (changes requested):

ktrushin (approved):

sandraromanchenko (approved):

radoslawszulgo (commented):
Great work! Ana, I like these changes!

radoslawszulgo (approved):

ktrushin (approved):

PR #912, opened 2024-08-23
PSMDB-1284 Doc: File copy based initial sync

nastena1606:
new file:   docs/initial-sync.md
	modified:   docs/psmdb-pro.md
	modified:   mkdocs-base.yml


docs/initial-sync.md
```diff
@@ -0,0 +1,26 @@
+# File copy based initial sync
+
+!!! admonition "Version added: [7.0.14-8](release_notes/7.0.14-8.md)"
+
+When a new member joins the replica set, it receives the data from the primary node via the initial sync.
```
igorsol: Initial sync source is not always the primary node. There is special algorithm which select initial sync source among all replset nodes.

docs/initial-sync.md
```diff
@@ -0,0 +1,26 @@
+# File copy based initial sync
+
+!!! admonition "Version added: [7.0.14-8](release_notes/7.0.14-8.md)"
+
+When a new member joins the replica set, it receives the data from the primary node via the initial sync. 
+
+The default initial sync method is logical, during which Percona Server for MongoDB clones all databases except the `local` database, builds all collection indexes, pulls oplog records and applies the changes to the data set. Read more about [logical initial sync in MongoDB documentation](https://www.mongodb.com/docs/manual/core/replica-set-sync/#logical-initial-sync-process).
+
+Starting with version 7.0.12, you can select **file copy based** as the initial sync method. This is the physical copying of the data files from source to target. This sync method is faster than logical, which is especially beneficial in heavy write environments. It speeds up cluster scaling and increases restore performance.
```
igorsol: It will likely be other version, not 7.0.12

docs/initial-sync.md
```diff
@@ -0,0 +1,26 @@
...
+
+File copy-based initial sync is implemented in the same way as in MongoDB Enterprise. For workflow and known limitations, refer to [MongoDB documentation](https://www.mongodb.com/docs/manual/core/replica-set-sync/#file-copy-based-initial-sync).
```
igorsol: I would suggest to change "implemented the same way" to "implemented in a similar way".
Ideally we need to distinguish between internal implementation and user-facing configuration and behavior. Internal implementation is definitely not the same.

igorsol (approved):
Looks good but exact versions should be reviewed everywhere when we are ready for release

docs/initial-sync.md
nastena1606: Updated

docs/initial-sync.md
nastena1606: @radoslawszulgo, do you think we need to add details on how FCBIS is implemented in PSMDB? Add a sync workflow for example?

docs/initial-sync.md
```diff
@@ -0,0 +1,26 @@
+# File copy based initial sync
+
+!!! admonition "Version added: [7.0.21-12](release_notes/7.0.21-12.md)"
+
+When a new member joins the replica set, it receives the data from the existing replica set node via the initial sync. 
+
+The default initial sync method is logical, during which Percona Server for MongoDB clones all databases except the `local` database, builds all collection indexes, pulls oplog records and applies the changes to the data set. Read more about [logical initial sync in MongoDB documentation](https://www.mongodb.com/docs/manual/core/replica-set-sync/#logical-initial-sync-process).
```
radoslawszulgo: I'd remove this paragraph. Optionally, move this till the end and shortly inform that there's the other (default) initial sync method.

docs/initial-sync.md
```diff
@@ -0,0 +1,26 @@
...
+
+Starting with version 7.0.21-12, you can select **file copy based** as the initial sync method. You must have the WiredTiger storage defined in Percona Server for MongoDB configuration. File copy based initial sync is the physical copying of the data files from source to target. This sync method is faster than logical, which is especially beneficial in heavy write environments. It speeds up cluster scaling and increases restore performance.
+
+File copy-based initial sync is implemented in a similar way as in MongoDB Advanced and has the same configuration options. For workflow and known limitations, refer to [MongoDB documentation](https://www.mongodb.com/docs/manual/core/replica-set-sync/#file-copy-based-initial-sync).
```
radoslawszulgo: At best we should review and have our own list of limitations rather than redirecting to the Upstream - if they fix something and remove the limitation it doesn't mean it's fixed in our case.

olexandr-havryliak (approved):

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+
+When a new member joins the replica set, it receives the data from the existing replica set node via the initial sync. 
+
+In [Percona Server for MongoDB Pro](psmdb-pro.md), you can choose a file copy-based initial sync for a new node. You must have WiredTiger defined as the storage.
+
+File copy-based initial sync method is a physical copying of the data files from the source to the target. It is much faster than the default [logical initial sync :octicons-link-external-16:](https://www.mongodb.com/docs/manual/core/replica-set-sync/#logical-initial-sync-process), which is especially beneficial in heavy write environments. Using this sync method speeds up cluster scaling and increases restore performance. 
+
+File copy-based initial sync implementation is compatible with that of MongoDB Advanced and has the same [configuration parameters](#configuration-parameters).
+
+To select the initial sync method, specify the following configuration in the configuration file for the target node:
```
igorsol: not sure, maybe following wording is better?
> ... specify the following __parameter__ in the configuration file ...

igorsol: or consider this:
> ... specify the __initial sync method__ in the configuration file ...

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+
+```yaml
+setParameter:
+  initialSyncMethod: fileCopyBased
+```
+
+You can only set this configuration at startup.
```
igorsol: consider this:
> This server parameter can only be set at startup

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Workflow
+
+When you start a new node for the replica set, the workflow is the following:
+
+1. The new node (also referred to as the target node) selects for the source node for the sync. This sync source is typically the node that responded first and has the passing configuration (e.g. it has WiredTiger set as the storage and the same arrangement of files and indexes as the target node.)
```
igorsol: In this part:
> selects for the source node for the sync

the first "for" is redundant

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+2. The target node opens a `$backupCursor` towards the sync source. The `$backupCursor` returns the list of files to copy and the timestamp of the oplog end in the metadata file.
+3. The file copy starts. During this process the target node lags behind the sync source as it remains operational and its data changes. The sync source node is periodically checked to ensure the time of the lag falls within the dined time.
+4. If the lag between the sync source differs and the target exceeds the defined threshold, the target node opens the `$backupCursorExtended` to retrieve the changes. Depending on the file copy duration, the target node can open `$backupCursorExtended` several times, limited by the maximum number of cycles (3 by default)
```
igorsol: > If the lag between the sync source differs and the target exceeds the defined threshold

The "diiffers" should be removed

igorsol: I think it would be better to say that target node "executes" $backupCursorExtend instead of "opens"

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+5. When the files are copied and the lag between the sync source and the target is acceptable, the target node closes the `$backupCursor`.
```
igorsol: Please replace '$backupCursor'  with "backup cursor".
It is more natural in this context.

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+2. The target node opens a `$backupCursor` towards the sync source. The `$backupCursor` returns the list of files to copy and the timestamp of the oplog end in the metadata file.
+3. The file copy starts. During this process the target node lags behind the sync source as it remains operational and its data changes. The sync source node is periodically checked to ensure the time of the lag falls within the dined time.
+4. If the lag between the sync source differs and the target exceeds the defined threshold, the target node opens the `$backupCursorExtended` to retrieve the changes. Depending on the file copy duration, the target node can open `$backupCursorExtended` several times, limited by the maximum number of cycles (3 by default)
+5. When the files are copied and the lag between the sync source and the target is acceptable, the target node closes the `$backupCursor`. 
+6. The node internally moves the downloaded files to the local `dbPath`, applies oplog on top, reconstructs timestamps to ensure data consistency.
```
igorsol: The __target__ node

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Configuration parameters
+
+These configuration parameters can be used to control the file-copy-based initial sync flow. You can set them only at startup.
```
igorsol: "file-copy-based" should be "file copy-based"

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+| `fileBasedInitialSyncMaxCyclesWithoutProgress` | integer | 3 | Specifies the max number of cycles to clone updates while the lag between the syncing node and the sync source is higher than `fileBasedInitialSyncMaxLagSec` |
+
+
+## Limitations
+
+Using file copy based initial sync has the following limitations:
+
+* Don't run backups on either sync source or syncing nodes
+* Don't write to the `local` database on the syncing node
+* You cannot use the same sync source for multiple syncing nodes
```
igorsol: Single node cannot be used as a sync source for more than one target node simultaneously because only one backup cursor can exist at any moment.

docs/psmdb-pro.md
```diff
@@ -15,6 +15,7 @@ Find the list of solutions available in Percona Server for MongoDB Pro builds:
 | Name                                | Version added | Description  | 
 | ----------------------------------- | ------------- | -------------
 | [FIPS support ](fips.md)| [7.0.4-2](release_notes/7.0.4-2.md) | FIPS mode provides a way to use FIPS-compliant encryption and run the Percona Server for MongoDB with the FIPS-140 certified library for OpenSSL. This helps customers meet minimum security requirements for cryptographic modules and testing in both hardware and software. |
+| [File copy based initial sync](initial-sync.md) | [7.0.21-12](release_notes/7.0.21-12.md) | File copy based initial sync is a new method of syncing data between nodes in a cluster. It is faster than the logical sync method as it copies physical files rather than cloning data. |
```
igorsol: Consider replacing "a new method" with "an additional method"

igorsol: "method of syncing data between nodes in a cluster" might be misleading. More technically correct way to say this is:
> method of doing initial sync for the new node in a replica set

igorsol (changes requested):

docs/initial-sync.md
nastena1606: I'd leave the active voice. Replaced configuration with server parameter.

docs/psmdb-pro.md
```diff
@@ -15,7 +15,8 @@ Find the list of solutions available in Percona Server for MongoDB Pro builds:
 | Name                                | Version added | Description  | 
 | ----------------------------------- | ------------- | -------------
 | [FIPS support ](fips.md)| [7.0.4-2](release_notes/7.0.4-2.md) | FIPS mode provides a way to use FIPS-compliant encryption and run the Percona Server for MongoDB with the FIPS-140 certified library for OpenSSL. This helps customers meet minimum security requirements for cryptographic modules and testing in both hardware and software. |
-| [File copy based initial sync](initial-sync.md) | [7.0.21-12](release_notes/7.0.21-12.md) | File copy based initial sync is an additional method of doing an initial sync for a new node in a replica set. It is faster than the logical sync method as it copies physical files rather than cloning data. |
+| [File copy-based initial sync](initial-sync.md) | [7.0.21-12](release_notes/7.0.21-12.md) | File copy based initial sync is an additional method of doing an initial sync for a new node in a replica set. It is faster than the logical sync method as it copies physical files rather than cloning data. |
```
radoslawszulgo:
```suggestion
| [File copy-based initial sync](initial-sync.md) | [7.0.21-12](release_notes/7.0.21-12.md) | File copy-based initial sync is an additional method of doing an initial sync for a new node in a replica set. It is faster for big databases than the logical sync method as it copies physical files rather than cloning data. |
```

igorsol (approved):

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
+# File copy based initial sync
+
+!!! admonition "Version added: [7.0.22-12](release_notes/7.0.22-12.md)"
+
+When a new member joins the replica set, it receives the data from the existing replica set node via the initial sync. 
+
+In [Percona Server for MongoDB Pro](psmdb-pro.md), you can choose a file copy-based initial sync for a new node. You must have WiredTiger defined as the storage.
```
radoslawszulgo: If user build on their own - it's not "Pro" - so we should generally say it's in PSMDB.
```suggestion
In Percona Server for MongoDB, you can choose a file copy-based initial sync for a new node. You must have WiredTiger defined as the storage.
```

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+
+File copy-based initial sync method is a physical copying of the data files from the source to the target. It is much faster than the default [logical initial sync :octicons-link-external-16:](https://www.mongodb.com/docs/manual/core/replica-set-sync/#logical-initial-sync-process), which is especially beneficial in heavy write environments. Using this sync method speeds up cluster scaling and increases restore performance.
```
radoslawszulgo:
```suggestion
The file copy-based initial sync method is a physical copying of the data files from the source to the target. It is much faster than the default [logical initial sync :octicons-link-external-16:](https://www.mongodb.com/docs/manual/core/replica-set-sync/#logical-initial-sync-process) for big datasets (500GB+), which is especially beneficial in heavy write environments. Using this initial sync method speeds up cluster scaling and increases restore performance. 
```

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+
+!!! admonition "Version added: [7.0.22-12](release_notes/7.0.22-12.md)"
+
+When a new member joins the replica set, it receives the data from the existing replica set node via the initial sync. 
+
+In [Percona Server for MongoDB Pro](psmdb-pro.md), you can choose a file copy-based initial sync for a new node. You must have WiredTiger defined as the storage.
+
+File copy-based initial sync method is a physical copying of the data files from the source to the target. It is much faster than the default [logical initial sync :octicons-link-external-16:](https://www.mongodb.com/docs/manual/core/replica-set-sync/#logical-initial-sync-process), which is especially beneficial in heavy write environments. Using this sync method speeds up cluster scaling and increases restore performance. 
+
+File copy-based initial sync implementation is compatible with that of MongoDB Advanced and has the same [configuration parameters](#configuration-parameters).
```
radoslawszulgo:
```suggestion
File copy-based initial sync implementation is compatible with that of MongoDB Enterprise Advanced and has the same [configuration parameters](#configuration-parameters).
```

radoslawszulgo:
I'd change the file name to fcbis-initial-sync.md or something. As there's also logical initial sync - the name of the file (and the URL) might be misleading.

docs/initial-sync.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+## Limitations
+
+Using file copy based initial sync has the following limitations:
+
+* Don't run backups on either sync source or syncing nodes
+* Don't write to the `local` database on the syncing node
+* You cannot use the same sync source for multiple target nodes simultaneously because only one backup cursor can exist at any moment.
+* If you're using encrypted storage, Percona Server for MongoDB applies the encryption key from the sync source node to secure the data on the syncing node.
```
radoslawszulgo: Please add initial sync doesn't work for in-memory storage engine

nastena1606: Percona Memory Engine?

docs/fcbis.md
```diff
@@ -0,0 +1,59 @@
+# File copy based initial sync
+
+!!! admonition "Version added: [7.0.22-12](release_notes/7.0.22-12.md)"
+
+When a new member joins the replica set, it receives the data from the existing replica set node via the initial sync. 
+
+In Percona Server for MongoDB, you can choose a file copy-based initial sync for a new node. You must have WiredTiger defined as the storage.
```
radoslawszulgo, line 7:
```suggestion
In Percona Server for MongoDB, you can choose a file copy-based initial sync for a new node. 
```

docs/fcbis.md
```diff
@@ -0,0 +1,59 @@
...
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| `initialSyncMethod` | string | logical | Specifies which method of initial sync to use. Valid options are: fileCopyBased, logical. |
+| `numInitialSyncAttempts` | integer | 10 | Number of attempts of attempts to make at replica set initial synchronization |
+| `numInitialSyncConnectAttempts` | integer | 10 | The number of attempts to select and connect to a valid sync source |
+| `fileBasedInitialSyncMaxLagSec` | integer | 300 | Specifies the max lag in seconds between the syncing node and the sync source to mark the file copy based initial sync as done successfully |
+| `fileBasedInitialSyncMaxCyclesWithoutProgress` | integer | 3 | Specifies the max number of cycles to clone updates while the lag between the syncing node and the sync source is higher than `fileBasedInitialSyncMaxLagSec` |
+
+
```
radoslawszulgo, line 45: I'd appreciate a full configuration example with all the options

nastena1606: As discussed with @igorsol we don't want users to mess with these params. yes, we list it. And the defaults should be enough. The way how to enable FCBIS is given.
@igorsol please confirm

igorsol: I agree that its is enough to list parameters with descriptions. By default FCBIS should work with default parameters. Only dbadmins with good understanding of the topic should try to alter those parameters. I assume they know how to put parameter into config file.

PR #928, opened 2024-09-19
PSMDB-1536 Updated upgrade doc to use percona-release

nastena1606:
modified:   docs/install/update-pro.md

### DIFF
--- a/docs/install/update-pro.md
+++ b/docs/install/update-pro.md
@@ -71,13 +71,19 @@ This document provides instructions how you can upgrade from Percona Server for 
 
         === ":octicons-file-code-24: Configuration file"            
 
-            Create the `/root/.percona-private-repos.config` configuration file with the following content:            
+            1. Create the `/root/.percona-private-repos.config` configuration file with the following content:            
 
-            ```ini title="/root/.percona-private-repos.config"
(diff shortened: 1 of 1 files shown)


docs/install/update-pro.md
```diff
@@ -8,85 +8,116 @@ This document provides instructions how you can upgrade from Percona Server for
...
+            ```             
+
+        === ":octicons-file-code-24: Configuration file"            
+
+            Create the `/root/.percona-private-repos.config` configuration file with the following content:            
+
+            ```ini title="/root/.percona-private-repos.config"
+            [psmdb-70-pro]
+            USER_NAME=<Your Customer ID>
+            REPO_TOKEN=<Your PRO repository token>
```
nastena1606: @radoslawszulgo, do you think it is feasible to remove this option? According to @sandraromanchenko we need to run the command percona-release enable psmdb-70-pro anyway. So the option to define creds in the config file seems to bring little value

radoslawszulgo: Yes, that's essential piece. Why do we wanted to remove it?

radoslawszulgo (approved):

olexandr-havryliak (approved):

PR #934, opened 2024-10-02
PSMDB-1441 Documented the key loss prevention

nastena1606:


docs/vault.md
```diff
@@ -35,9 +39,27 @@ security:
...
+}
+path "secret/config" {
+  capabilities = ["read"]
+}
+```
+
+During the first run of the Percona Server for MongoDB, the process generates a new random master encryption key. Then, it wraps it into a secret and puts the latter on a Vault server at the configured path. Vault increments the value of the `current_version`, associates the resulting value with a new secret, and returns the version. Percona Server for MongoDB then saves the full path and the version in the metadata and uses them later to get the key from the Vault server.
 
-During the subsequent start, the server tries to read the master key from the vault. If the configured secret does not exist, vault responds with HTTP 404 error.
+During the subsequent start, the server tries to read the master key from the vault. If the configured secret does not exist, vault responds with the HTTP 404 error.
```
ktrushin: Do we need a capital letter for Vault? I mean `key from the Vault` and `not exist, Vault responds`.

docs/vault.md
```diff
@@ -90,11 +112,24 @@ Rotating the master key process also re-encrypts the keystore using the new mast
 
 ### Key rotation in replica sets
 
-Starting with version [6.0.5-4](https://docs.percona.com/percona-server-for-mongodb/6.0/release_notes/6.0.5-4.md), you can store the master key at the same path on every replica set member in your entire deployment. Vault assigns different versions to the master keys stored at the same path. The path and the version serve as the unique identifier of a master key. The `mongod` server stores that identifier and uses it to retrieve the correct master key from the Vault server during the restart.  
+Starting with version [6.0.5-4](https://docs.percona.com/percona-server-for-mongodb/6.0/release_notes/6.0.5-4.md), you can store the master key at the same path on every replica set member in your entire deployment. Vault assigns different versions to the master keys stored at the same path. The path and the version serve as the unique identifier of a master key. The `mongod` server stores that identifier and uses it to retrieve the correct master key from the Vault server during the restart.
```
ktrushin: I see unnecessary spaces after the last dot: `during the restart.   ` Can they be removed?

ktrushin (approved):
Looks good except of a pair of minor typos. When corrected, there is no need for another review cycle. Please proceed to the merging.

PR #958, opened 2025-01-14
PSMDB-1591 Moved Get help banner from bottom to the right

nastena1606:
Added an extra Get more help page
Updated a GitHub action


alina-derkach-oaza:
the old banner is still at the bottom of the page

alina-derkach-oaza (approved):

PR #1066, opened 2025-10-08
PSMDB-1821 Updated Docker steps 

nastena1606:
With the consideration about the restart policy if using PBM for physical backups and restores

modified:   docs/install/docker.md


igroene (approved):
LGTM

olexandr-havryliak (approved):

PR #1071, opened 2025-10-15
Remove leading dollar signs from command examples

radoslawszulgo:
Updated command examples to remove unnecessary leading dollar signs.


nastena1606:
Updating a single page breaks consistency all over this doc set and others. I checked and $ are removed when clicking the copy button. 
This must be a global decision to proceed with or without leading $.

radoslawszulgo:
Ok. I just used the copy button and it included the dollar sign

PR #1089, opened 2025-12-24
PSMDB-1902 Fixed command rendering

nastena1606:


docs/sasl-auth.md
```diff
@@ -72,15 +72,7 @@ Before we move on to the configuration steps, we assume the following:
          Use the following commands to enable the `saslauthd` to auto-run on startup and to set the `ldap` value for the `--MECHANISMS` option:
 
          ```bash
-         sudo sed -i -e s/^MECH=pam/MECH=ldap/g /etc/sysconfig/saslauthdsudo sed -i -e s/^MECHANISMS="pam"/MECHANISMS="ldap"/g /etc/default/saslauthd
```
igorsol, line 75: This is definitely a mess: in fact it seems there was two lines:
```
sudo sed -i -e s/^MECH=pam/MECH=ldap/g /etc/sysconfig/saslauthd
sudo sed -i -e s/^MECHANISMS="pam"/MECHANISMS="ldap"/g /etc/default/saslauthd
```
The first line seems to be intended for Red Hat-based Linux distributions because `/etc/sysconfig` does not exists on Debian-based distributions.
Second line looks OK for Debian-based systems

docs/sasl-auth.md
```diff
@@ -72,15 +72,7 @@ Before we move on to the configuration steps, we assume the following:
...
-         sudo sed -i -e s/^START=no/START=yes/g /etc/default/saslauthd
```
igorsol, line 76: This line refers to Debian-specific file path `/etc/default/saslauthd` however that file has no `START` key so this `sed` invocation is just a no-op.

docs/sasl-auth.md
```diff
@@ -72,15 +72,7 @@ Before we move on to the configuration steps, we assume the following:
...
-         ```
-
-         Alternatively, you can edit the `/etc/default/sysconfig/saslauthd` configuration file:
-
-         ```init
-         START=yes
-         MECHANISMS="ldap
+         sudo sed -i -e "s/START=.*/START=yes/" -e "s/MECHANISMS=.*/MECHANISMS=\"ldap\"/" /etc/default/saslauthd
          ```
```
igorsol, line 82: My guess is that `START` and `MECHANISMS` are in two different files.

docs/sasl-auth.md
```diff
@@ -72,15 +72,7 @@ Before we move on to the configuration steps, we assume the following:
...
-         sudo sed -i -e s/^MECH=pam/MECH=ldap/g /etc/sysconfig/saslauthdsudo sed -i -e s/^MECHANISMS="pam"/MECHANISMS="ldap"/g /etc/default/saslauthd
-         sudo sed -i -e s/^START=no/START=yes/g /etc/default/saslauthd
-         ```
-
-         Alternatively, you can edit the `/etc/default/sysconfig/saslauthd` configuration file:
-
-         ```init
-         START=yes
-         MECHANISMS="ldap
+         sudo sed -i -e "s/START=.*/START=yes/" -e "s/MECHANISMS=.*/MECHANISMS=\"ldap\"/" /etc/default/saslauthd
```
igorsol: The `START` does not exist in `/etc/default/saslauthd`.
We need to identify where it should go.
It seems we also need to cover differences between Debian and RHEL

igorsol (changes requested):

docs/sasl-auth.md
igorsol: I will try to find more information on this and will comment here.

igorsol: On debian-based systems we need this to set "ldap" mechanism:
```
sudo sed -ri s/^MECHANISMS=.*$/MECHANISMS=\"ldap\"/g /etc/default/saslauthd
```
On redhat-based systems this:
```
sudo sed -ri -e s/^MECH=.*$/MECH=ldap/g /etc/sysconfig/saslauthd
```

igorsol: The START parameter has no effect on debian-based systems so no need to set or change it.
On debian -based systems the `saslauthd` service is not configured to auto start by default - to enable auto start we need to execute this command:
```
sudo systemctl enable saslauthd
```

I have no information about START behavior on redhat-based systems

docs/sasl-auth.md
```diff
@@ -69,18 +69,16 @@ Before we move on to the configuration steps, we assume the following:
...
+         Use the following command to set the `ldap` value for the `--MECHANISMS` option:
 
          ```bash
-         sudo sed -i -e s/^MECH=pam/MECH=ldap/g /etc/sysconfig/saslauthdsudo sed -i -e s/^MECHANISMS="pam"/MECHANISMS="ldap"/g /etc/default/saslauthd
-         sudo sed -i -e s/^START=no/START=yes/g /etc/default/saslauthd
+         sudo sed -ri s/^MECHANISMS=.*$/MECHANISMS=\"ldap\"/g /etc/default/saslauthd
          ```
 
-         Alternatively, you can edit the `/etc/default/sysconfig/saslauthd` configuration file:
+         Configure the `saslauthd` service to auto-start at startup:
```
sandraromanchenko: This is duplicate from step below in doc "Start the saslauthd process and set it to run at restart", no need to repeat, leave the same note as for RHEL

```
Alternatively, you can edit the /etc/default/saslauthd configuration file:
MECHANISMS="ldap"
```

sandraromanchenko (changes requested):

sandraromanchenko (approved):

docs/sasl-auth.md
```diff
@@ -69,18 +69,16 @@ Before we move on to the configuration steps, we assume the following:
     
     === ":material-debian: Debian and Ubuntu"
 
-         Use the following commands to enable the `saslauthd` to auto-run on startup and to set the `ldap` value for the `--MECHANISMS` option:
+         Use the following command to set the `ldap` value for the `--MECHANISMS` option:
```
igorsol: There should be no dash-dash prefix before MECHANISMS.
The same is also true for the --MECH in RHEL section

igorsol (changes requested):

igorsol (approved):

PR #1117, opened 2026-04-15
Apply-new-branding-(colors,-icons)

rasika-chivate:


alina-derkach-oaza:
Close this one. We do not need to update that file in the templates folder

PR #1125, opened 2026-05-12
Add Percona Live banner to docs

rasika-chivate:
Add Percona Live banner to docs


patrickbirch (approved):
k

radoslawszulgo (approved):
Shouldn't we use "open-source"?

rasika-chivate:
@radoslawszulgo We used it without the "-" for all the websites. Let's use that convention only.

PR #1158, opened 2026-07-31
PSMDB-2150 SBOMs

rasika-chivate:
Include SBOM into PSMDB binary packages

### DIFF
--- a/docs/sbom.md
+++ b/docs/sbom.md
@@ -1,0 +1,139 @@
+# Software Bill of Materials
+
+A Software Bill of Materials (SBOM) is a machine-readable inventory of the components and dependencies included in a software release. It helps you understand what is included in a build and assess potential security or compliance risks.
+
+Starting with version 7.0.39-21, every Percona Server for MongoDB (PSMDB) release includes a [CycloneDX :octicons-link-external-16:](https://cyclonedx.org/specification/overview/){:target="_blank"} SBOM in JSON format.
+
+## Why it matters
+
+An SBOM helps you:
+
+- Identify the components and dependencies included in a PSMDB release.
+- Assess known vulnerabilities using SBOM-compatible security scanners.
+- Support security reviews, compliance processes, and software supply chain requirements.
+- Verify the contents of deployed software artifacts.
+
+## Where to find the SBOM
+
+| Distribution method | SBOM location |
+|---|---|
+| Binary tarball | `doc/sbom.cdx.json` |
+| RPM package | `/usr/share/doc/percona-server-mongodb-server/sbom.cdx.json` |
+| DEB package | `/usr/share/doc/percona-server-mongodb-server/sbom.cdx.json` |
+| Docker image | Embedded in the image and available as an attached OCI artifact. See [Docker images](#docker-images). |
+
+
+## Verifying and scanning the SBOM
+
+The examples below use [Grype :octicons-link-external-16:](https://github.com/anchore/grype){:target="_blank"}.
+
+!!! note
+    [Trivy :octicons-link-external-16:](https://trivy.dev/){:target="_blank"} cannot currently scan the SBOMs included with Percona Server for MongoDB DEB and RPM packages or binary tarballs. Most dependencies in these SBOMs are identified using the GitHub package type, which Trivy does not fully support in this context.
+
+    Trivy can, however, scan the SBOMs attached to Percona Server for MongoDB Docker images as OCI artifacts.
+
+### Binary tarball
+
+```bash
+# Confirm the SBOM is bundled
+tar tzf percona-server-mongodb-{{ release }}-x86_64.<os_codename>.tar.gz \
+    | grep doc/sbom.cdx.json
+
+# Extract and scan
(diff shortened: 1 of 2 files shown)


rasika-chivate:
@ktrushin As discussed, I've not added any information for OpenVEX files. Please confirm if we need to add this.

ktrushin (approved):

docs/sbom.md
```diff
@@ -0,0 +1,139 @@
...
+### DEB package
+
+```bash
+# Confirm the package installs the SBOM
+dpkg -L percona-server-mongodb-server | grep sbom.cdx.json
+
+# Scan it (replace `ubuntu:24.04` with your `<os_name>:<os_version>`)
+grype --distro ubuntu:24.04 sbom:/usr/share/doc/percona-server-mongodb-server/sbom.cdx.json
+```
+
```
ktrushin, line 67: We can also document using VEX files here. Below is the draft version of the suggested change:
``````
### Using the VEX File
One can also utilize the VEX file located at https://percona.github.io/percona-server-mongodb/vex/percona-server-mongodb-{{ release }}.vex.json to filter out vulnerabilities that do not affect this particular Percona Server for MongoDB version:
```bash
curl -fsSLO https://percona.github.io/percona-server-mongodb/vex/percona-server-mongodb-{{ release }}.vex.json
grype --distro <os_name>:<os_version> --vex=percona-server-mongodb-{{ release }}.vex.json sbom:/usr/share/doc/percona-server-mongodb-server/sbom.cdx.json
```

Note. At the time of writing, the example above only produces CVE-2026-6231, which only affects Percona Server for MongoDB if it uses Querable Encryption aka Field Level Encryption. This vulnerability has existed for a while rather then having been introduced in version {{ release }}, which only makes it visible via SBOM. We plan to fix the vulnerability in a future release.

``````

Feel free to change the wording for better clarity if required. Also, input from @radoslawszulgo may be helpful.

File: docs/sbom.md [hunk 2]
```diff
@@ -136,4 +136,32 @@ Follow these steps:
...
         percona-server-mongodb-{{ release }}-amd64.cdx.json
-        ```
\ No newline at end of file
+        ```
+
+## Filtering vulnerabilities with OpenVEX
+
+[OpenVEX :octicons-link-external-16:](https://github.com/openvex/spec){:target="_blank"} is an open, minimal format for Vulnerability Exploitability eXchange (VEX) statements. A VEX document records whether a known vulnerability (CVE) in a component actually affects a given product — for example, the vulnerable code path is unreachable, the issue is already fixed, or it's still under investigation. Scanners that support VEX use these statements to suppress vulnerabilities that don't actually apply, cutting down on noise from CVEs in bundled dependencies that PSMDB isn't actually exposed to.
+
+Percona publishes an OpenVEX document for every PSMDB release, plus one combined document covering all releases and CVE statuses:
```
ktrushin: > , plus one combined document covering all releases and CVE statuses:

As discussed [here](https://github.com/percona/percona-server-mongodb/pull/2083#discussion_r3704072672), that "all realizes" VEX file can't work reliably and is better removed from the documentation.

@radoslawszulgo could you please confirm that?

ktrushin (changes requested):

docs/sbom.md
radoslawszulgo: confirm.  Let's remove these references in docs.

docs/sbom.md
radoslawszulgo: I think we've documented that in openvex section below?

docs/sbom.md
```diff
@@ -0,0 +1,162 @@
...
+
+[OpenVEX :octicons-link-external-16:](https://github.com/openvex/spec){:target="_blank"} is an open, minimal format for Vulnerability Exploitability eXchange (VEX) statements. A VEX document records whether a known vulnerability (CVE) in a component actually affects a given product — for example, the vulnerable code path is unreachable, the issue is already fixed, or it's still under investigation. Scanners that support VEX use these statements to suppress vulnerabilities that don't actually apply, cutting down on noise from CVEs in bundled dependencies that PSMDB isn't actually exposed to.
+
+Percona publishes an OpenVEX document for every PSMDB release.
+
+The `VEX` file for version {{ release }} is available at:
+
+`https://percona.github.io/percona-server-mongodb/vex/percona-server-mongodb-{{ release }}.openvex.json`
+
+Pass the document to Grype's `--vex` flag together with any SBOM scan on this page. For example, combined with the binary tarball scan:
```
ktrushin: Actually, the example below uses SBOM path from DEB/RPM package.
I suggest changing the phrase as below:

```suggestion
Pass the document to Grype's `--vex` flag together with any SBOM scan on this page. For example, combined with scanning an RPM package's SBOM:
```

ktrushin (approved):
Looks good except for the minor inaccuracy. Feel free to fix it the way you think would be the best. No need in another review cycle.

docs/sbom.md
rasika-chivate: Reworded to "Pass the document to Grype's --vex flag along with any SBOM scan on this page. For example, when scanning the SBOM of an RPM package:"

PR #1175, opened 2026-09-01
Update fcbis.md

vinokrishna:
need some doc update to include that FCBIS in v8.0.12-4 is available with Probuild or need to be compiled from the source.


igroene:
pro builds removed in 8.0.16 so let's mention any version older than 8.0.16 also for 7.0 branch we have to do similar fix

PR #1176, opened 2026-09-01
Change requestScopes to array format in oidc-entra.md

Starfox64:
I copy pasted the example in a test cluster and it failed because an array was expected.


radoslawszulgo (approved):
Thanks for the PR!

